PR #5, opened 2024-04-04
feat: remove template stuff and db connection

lauti7:


package.json
```diff
@@ -1,13 +1,15 @@
 {
-  "name": "template-server",
+  "name": "social-service-ea",
   "scripts": {
     "build": "tsc -p tsconfig.json",
     "start": "node --trace-warnings --abort-on-uncaught-exception --unhandled-rejections=strict dist/index.js",
-    "test": "jest --forceExit --detectOpenHandles --coverage --verbose"
+    "test": "jest --forceExit --detectOpenHandles --coverage --verbose",
+    "migrate": "node-pg-migrate --database-url-var PG_COMPONENT_PSQL_CONNECTION_STRING --envPath .env -j ts --tsconfig tsconfig.json -m ./src/migrations"
```
marianogoldman, line 7: I like this approach more than what we have in other apps (like worlds content server), this way we have to be explicit about when we want to run the migration. However... how are we going to handle this when deploying to dev/prd?
In WCS we run migration during app startup, so you don't need to separate steps, so just deploying the up and starting does the trick.

src/adapters/db.ts
```diff
@@ -0,0 +1,7 @@
+import { AppComponents } from '../types'
+
+export interface IDBComponent {}
```
marianogoldman, line 3:
```suggestion
export interface IDatabaseComponent {}
```

marianogoldman:
You need to add a license file and declare it in package.json

marianogoldman (approved):

PR #11, opened 2024-04-23
feat: friendships

lauti7:
This PR:
- updates an existing migration and adds a new table migration
- adds Redis to leverage the Pub/Sub system
- adds DB queries to handle and manage friendships
- adds business logic for friendships
- removes local protocol and uses the .proto from `@dcl/protocol`


src/ports/redis.ts
```diff
@@ -0,0 +1,37 @@
+import { createClient } from 'redis'
```
hugoArregui, line 1: we don't use ports directory anymore, they are called `adapters` now

hugoArregui (approved):
I think in general is ok, I don't fully understand how is redis been used

PR #13, opened 2024-04-26
feat: integration tests

lauti7:


src/adapters/db.ts
```diff
@@ -195,16 +193,20 @@ export function createDBComponent(components: Pick<AppComponents, 'pg' | 'logs'>
...
-        return res
+        res = callbackResult
       } catch (error) {
         logger.error(error as any)
         await client.query('ROLLBACK')
         client.release()
         throw error
+      } finally {
+        client.release()
+        return res as T
```
hugoArregui, line 209: you can return from the try cause and the finally will be executed anyway

src/adapters/db.ts
```diff
@@ -193,20 +193,18 @@ export function createDBComponent(components: Pick<AppComponents, 'pg' | 'logs'>
...
       try {
-        const callbackResult = await cb(client)
+        const res = await cb(client)
         await client.query('COMMIT')
-        res = callbackResult
+        return res
       } catch (error) {
         logger.error(error as any)
         await client.query('ROLLBACK')
         client.release()
```
hugoArregui, line 204: not sure if this a problem, but you are releasing it twice, once here and one one finally

.github/workflows/build.yml
```diff
@@ -1,8 +1,62 @@
-name: build
+name: build-app
 
 on:
   push:
 
 jobs:
-  build:
-    uses: decentraland/platform-actions/.github/workflows/apps-with-db-build.yml@main
+  validations:
```
marianogoldman, line 7: Why this? What's missing on the `apps-with-db-build.yml`?

PR #25, opened 2025-01-07
feat: Add missing indexes and constraints

kevinszuchet:


src/migrations/1736277138587_add-indexes.ts
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
braianj, line 1: no needed

braianj (approved):
👍🏽

PR #31, opened 2025-01-13
refactor: RPC services to different files + tests

kevinszuchet:


src/adapters/rpc-server/services/get-mutual-friends.ts
```diff
@@ -0,0 +1,45 @@
...
+import { normalizeAddress } from '../../../utils/address'
+
+export function getMutualFriendsService({ components: { logs, db } }: RPCServiceContext<'logs' | 'db'>) {
+  const logger = logs.getLogger('get-mutual-friends-service')
+
+  return async function* (request: MutualFriendsPayload, context: RpcServerContext): AsyncGenerator<UsersResponse> {
+    logger.debug(`getting mutual friends ${context.address}<>${request.user!.address}`)
+    let mutualFriends: AsyncGenerator<{ address: string }> | undefined
+    try {
+      mutualFriends = db.getMutualFriends(context.address, normalizeAddress(request.user!.address))
```
aleortega, line 16: Only second address needs to be normalized?

aleortega: What about using LOWER() Postgre's built-in function for this instead of enforcing lower cases during execution?

aleortega (approved):
LGTM!

src/adapters/rpc-server/services/get-mutual-friends.ts
kevinszuchet: The address that comes from the context is already normalized. The ws server injects it in the context of each of the RPC services. I'll use pg LOWER.

PR #32, opened 2025-01-16
feat: Implement new protocol definition

kevinszuchet:


src/adapters/rpc-server/services/get-pending-friendship-requests.ts
```diff
@@ -1,13 +1,18 @@
...
+} from '@dcl/protocol/out-ts/decentraland/social_service/v3/social_service_v3.gen'
 
 export function getPendingFriendshipRequestsService({ components: { logs, db } }: RPCServiceContext<'logs' | 'db'>) {
   const logger = logs.getLogger('get-pending-friendship-requests-service')
 
-  return async function (_request: Empty, context: RpcServerContext): Promise<FriendshipRequestsResponse> {
+  return async function (
+    request: GetFriendshipRequestsPayload,
+    context: RpcServerContext
+  ): Promise<PaginatedFriendshipRequestsResponse> {
```
aleortega, line 13: NIT: I'm thinking that maybe we could wrap-up pagination functionality in something like:

`Paginated<FriendshipRequestResponse>`

And so on for all the things we do need to paginate, what do you think?

src/adapters/rpc-server/services/upsert-friendship.ts
```diff
@@ -2,7 +2,7 @@ import { Action, FriendshipStatus, RpcServerContext, RPCServiceContext } from '.
 import {
   UpsertFriendshipPayload,
   UpsertFriendshipResponse
-} from '@dcl/protocol/out-ts/decentraland/social_service_v2/social_service.gen'
+} from '@dcl/protocol/out-ts/decentraland/social_service/v3/social_service_v3.gen'
```
aleortega, line 5: Doubt, is it okay to also have the v3 at the file name level and not only at route level?

src/adapters/db.ts
```diff
@@ -28,21 +54,33 @@ export function createDBComponent(components: Pick<AppComponents, 'pg' | 'logs'>
   const logger = logs.getLogger('db-component')
 
   return {
-    getFriends(userAddress, onlyActive = true) {
-      let query: SQLStatement
+    async getFriends(userAddress, { onlyActive, pagination } = { onlyActive: true }) {
+      const { limit, offset } = getPaginationOrDefaults(pagination)
```
aleortega, line 58: What about placing the default directly on this `getFriends` function's params instead at `getPaginationOrDefaults` as we do with `onlyActive:true`

aleortega: Also, could be possible that limit is set but not the offset, or vice versa?

aleortega (approved):
LGTM, left some NITS and doubts.

src/adapters/rpc-server/services/get-friendship-status.ts
```diff
@@ -0,0 +1,61 @@
...
+    request: GetFriendshipStatusPayload,
+    context: RpcServerContext
+  ): Promise<GetFriendshipStatusResponse> {
+    try {
+      const lastFriendshipAction = await db.getLastFriendshipActionByUsers([context.address, request.user!.address])
+
+      if (!lastFriendshipAction) {
+        return {
+          response: {
+            $case: 'internalServerError',
```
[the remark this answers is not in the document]
kevinszuchet: It came from the previous definition. We can rethink the error responses.

src/adapters/rpc-server/services/get-pending-friendship-requests.ts
kevinszuchet: In terms of the functionality, I agree, but the types come from the protocol, which doesn't allowtheo create generics

src/adapters/rpc-server/services/upsert-friendship.ts
kevinszuchet: Good catch

PR #39, opened 2025-01-23
feat: Send profile info on upsert and friendship updates

kevinszuchet:


src/adapters/rpc-server/services/upsert-friendship.ts
```diff
@@ -33,20 +36,9 @@ export function upsertFriendshipService({
...
+      const lastAction = await db.getLastFriendshipActionByUsers(context.address, parsedRequest.user!)
 
-      if (
-        !validateNewFriendshipAction(
-          context.address,
-          { action: parsedRequest.action, user: parsedRequest.user },
-          lastAction
-        )
-      ) {
+      if (!validateNewFriendshipAction(context.address, parsedRequest, lastAction)) {
```
aleortega, line 41: Shouldn't this function also receive the new action the user is trying?

src/adapters/peer-tracking.ts
```diff
@@ -56,10 +56,16 @@ export function createPeerTrackingComponent({
   }
 
   return {
-    async start() {
```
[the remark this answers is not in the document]
kevinszuchet: The start is optional, and it was a patch to wait first for the NATS component to be ready 🙃

src/adapters/rpc-server/services/upsert-friendship.ts
kevinszuchet: It is inside the parsed request

PR #48, opened 2025-02-03
feat: Enhance ws errors and reconnections

kevinszuchet:


src/adapters/ws-pool.ts
```diff
@@ -0,0 +1,109 @@
+import { AppComponents, IWSPoolComponent } from '../types'
+
+export async function createWSPoolComponent({
+  metrics,
+  config,
+  redis
+}: Pick<AppComponents, 'metrics' | 'config' | 'redis'>): Promise<IWSPoolComponent> {
+  const maxConcurrentConnections = (await config.getNumber('MAX_CONCURRENT_CONNECTIONS')) || 100
+  const idleTimeout = (await config.getNumber('IDLE_TIMEOUT')) || 300000 // 5 minutes default
```
aleortega, line 9:
```suggestion
  const idleTimeoutInMs = (await config.getNumber('IDLE_TIMEOUT_IN_MS')) || 300000 // 5 minutes default
```

src/adapters/ws-pool.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  const cleanupInterval = setInterval(async () => {
+    const now = Date.now()
+    const pattern = 'ws:conn:*'
+
+    for await (const key of redis.client.scanIterator({ MATCH: pattern })) {
+      const data = await redis.client.get(key)
+      if (data) {
+        const parsed = JSON.parse(data)
+        if (now - parsed.lastActivity > idleTimeout) {
```
aleortega, line 19: This is a safe check, right? because the TTL set at redis should already be enough to prevent this but maybe it is worth leaving it here

src/adapters/ws-pool.ts
```diff
@@ -0,0 +1,109 @@
...
+        .zAdd('ws:active_connections', { score: Date.now(), value: id })
+        .exec()
+
+      if (!result) {
+        throw new Error('Transaction failed')
+      }
+
+      const [totalConnections] = result as [number]
+
+      if (totalConnections && totalConnections >= maxConcurrentConnections) {
```
aleortega, line 55: Does this mean that no more than 100 users can connect/subscribe to this WS? is it safe to have it? how will we scale?

src/controllers/handlers/ws-handler.ts
```diff
@@ -3,95 +3,177 @@ import { onRequestEnd, onRequestStart } from '@well-known-components/uws-http-se
...
 
-  function changeStage(data: WsUserData, newData: WsUserData) {
-    Object.assign(data, newData)
+  function changeStage(data: WsUserData, newData: Partial<WsUserData>) {
+    Object.assign(data, { ...data, ...newData })
   }
 
   server.app.ws<WsUserData>('/', {
-    idleTimeout: 0,
+    idleTimeout: (await config.getNumber('WS_IDLE_TIMEOUT')) ?? 90, // In seconds
```
aleortega, line 23:
```suggestion
    idleTimeout: (await config.getNumber('WS_IDLE_TIMEOUT_IN_SECONDS')) ?? 90, // In seconds
```

src/controllers/handlers/ws-handler.ts
```diff
@@ -3,95 +3,177 @@ import { onRequestEnd, onRequestStart } from '@well-known-components/uws-http-se
...
+          data.eventEmitter.emit('close', { code, reason: messageText })
+          data.eventEmitter.all.clear()
+        } catch (error: any) {
+          logger.error('Error during connection cleanup', {
+            error: error.message,
+            address: data.address,
+            clientId
+          })
+        }
+        metrics.increment('ws_connections', { address: data.address })
```
aleortega, line 152: shouldn't we decrement on this `close` function?

src/controllers/handlers/ws-handler.ts
```diff
@@ -3,95 +3,177 @@ import { onRequestEnd, onRequestStart } from '@well-known-components/uws-http-se
...
+      } else {
+        metrics.increment('ws_connections')
```
aleortega, line 154: Maybe these metrics should be triggered by the WS Pool so clients can know nothing about them?

src/logic/notifications.ts
```diff
@@ -110,6 +110,5 @@ export async function sendNotification(
       senderAddress: context.senderAddress,
       receiverAddress: context.receiverAddress
     })
-    throw error
```
aleortega, line 113: What would happen now if we fail to send a notification? will it be retried later or we will show an error message on client side? how is that handle?

src/utils/UWebSocketTransport.ts
```diff
@@ -28,66 +30,154 @@ export interface IUWebSocket<T extends { isConnected: boolean }> {
...
-
-  function flush() {
-    for (const $ of queue) {
-      send($)
-      queue.length = 0
+  uServerEmitter: Emitter<IUWebSocketEventMap>,
+  config: IConfigComponent
+): Promise<Transport> {
+  const maxQueueSize = (await config.getNumber('WS_TRANSPORT_MAX_QUEUE_SIZE')) || 1000
+  const queueDrainTimeout = (await config.getNumber('WS_TRANSPORT_QUEUE_DRAIN_TIMEOUT')) || 5000
```
aleortega, line 39:
```suggestion
  const queueDrainTimeoutInMs = (await config.getNumber('WS_TRANSPORT_QUEUE_DRAIN_TIMEOUT_IN_MS')) || 5000
```

src/adapters/ws-pool.ts
kevinszuchet: Yep, it is just a proactive way of cleaning up idle connections. Redis will eventually clean this up even without this mechanism.

src/adapters/ws-pool.ts
kevinszuchet: I removed the limit

src/controllers/handlers/ws-handler.ts
```diff
@@ -3,95 +3,172 @@ import { onRequestEnd, onRequestStart } from '@well-known-components/uws-http-se
...
+            error,
+            address: data.address,
+            clientId: data.clientId,
+            isConnected: String(data.isConnected),
+            hasEventEmitter: String(!!data.eventEmitter)
+          })
+          metrics.increment('ws_errors', { address: data.address })
+          ws.send(JSON.stringify({ error: 'Error processing message', message: error.message }))
+        }
+      } else if (isNotAuthenticated(data)) {
```
aleortega, line 86: Is there any different state besides `data.auth` or `isNotAuthenticated`?

src/metrics.ts
```diff
@@ -1,10 +1,46 @@
...
+  },
+  ws_auth_errors: {
+    type: IMetricsComponent.CounterType,
+    help: 'Number of WebSocket authentication errors',
+    labelNames: ['address']
+  },
+  ws_idle_timeouts: {
+    type: IMetricsComponent.CounterType,
+    help: 'Number of WebSocket idle timeouts',
+    labelNames: ['id']
```
aleortega, line 42: Can we confirm with data team that adding this kind of labels to CounterType metrics won't cause any issues regarding cardinality?

src/types.ts
```diff
@@ -180,12 +191,15 @@ export type WsAuthenticatedUserData = {
   eventEmitter: Emitter<IUWebSocketEventMap>
   auth: true
   address: string
+  clientId: string
```
aleortega, line 194: Just curious, what's client id on this context and also on `WsNotAuthenticatedUserData`

test/mocks/components/metrics.ts
```diff
@@ -0,0 +1,11 @@
+import { IMetricsComponent } from '@well-known-components/interfaces'
+
+export const mockMetrics: jest.Mocked<IMetricsComponent<string>> = {
```
aleortega, line 3: Just a heads up about the existence of `createTestMetricsComponent` (_from WKC/metrics package_) maybe is already enough for these suites :)

test/unit/adapters/rpc-server.spec.ts
```diff
@@ -68,12 +78,54 @@ describe('createRpcServerComponent', () => {
...
       rpcServer.attachUser({ transport: mockTransport, address })
 
       expect(mockTransport.on).toHaveBeenCalledWith('close', expect.any(Function))
+      expect(attachTransportMock).toHaveBeenCalledWith(mockTransport, {
+        subscribers: expect.any(Object),
+        address
+      })
+    })
+
+    it('should clean up subscribers when transport closes', () => {
```
aleortega, line 92: can we add a expect about this call or the only way to testing is through `rpcServer.detachUser` current behavior?

test/unit/adapters/rpc-server.spec.ts
```diff
@@ -68,12 +78,54 @@ describe('createRpcServerComponent', () => {
...
+
+  describe('detachUser', () => {
+    const address = '0x123'
+
+    beforeEach(() => {
+      // Setup: attach a user first
+      rpcServer.attachUser({ transport: mockTransport, address })
+    })
+
+    it('should clean up subscriber when detaching user', () => {
```
aleortega, line 117: maybe?:
```suggestion
    it('should not throw when detaching an already detached user', () => {
```

test/unit/adapters/rpc-server.spec.ts
aleortega: or add a test with its counterpart:

`should throw when detaching an user from RPC if subscriber was not priorly cleaned`

test/unit/adapters/ws-pool.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      const pool = await createWSPoolComponent({ metrics: mockMetrics, config: mockConfig, redis: mockRedis })
+      expect(pool).toBeDefined()
+    })
+
+    it('should use configured idle timeout', async () => {
+      const customTimeout = 60000
+      mockConfig.getNumber.mockImplementation(async (key) => (key === 'IDLE_TIMEOUT_IN_MS' ? customTimeout : 0))
+      const pool = await createWSPoolComponent({ metrics: mockMetrics, config: mockConfig, redis: mockRedis })
+      expect(pool).toBeDefined()
+    })
```
aleortega, line 61: I'm not sure if we really want to put efforts on testing this 😆 but if we do really want to, should we add an `expect(redisMock.set).toHaveBeenCalledWith({...})` after calling `acquireConnection`?

test/unit/adapters/ws-pool.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    it('should update connection activity', async () => {
+      const key = 'ws:conn:test-1'
+      await wsPool.updateActivity('test-1')
+
+      expect(mockRedis.put).toHaveBeenCalledWith(
+        key,
+        { lastActivity: expect.any(Number) },
+        {
+          XX: true,
+          EX: Math.ceil(300000 / 1000)
```
aleortega, line 174: Oh this seems a way of testing timeout config :)

src/types.ts
kevinszuchet: Renamed to `wsConnectionId`

aleortega (approved):

PR #49, opened 2025-02-04
feat: Fail fast on self friendship reqs

kevinszuchet:


aleortega (approved):

src/adapters/rpc-server/services/upsert-friendship.ts
```diff
@@ -34,7 +34,17 @@ export async function upsertFriendshipService({
       }
     }
 
-    logger.debug(`upsert friendship > `, parsedRequest as Record<string, string>)
```
aleortega, line 37: Shall we keep this log or is it no longer needed?

aleortega (approved):

PR #50, opened 2025-02-05
feat: Send more info in the notifications

kevinszuchet:
- [ ] Bump schemas


src/adapters/rpc-server/services/upsert-friendship.ts
```diff
@@ -96,7 +96,7 @@ export async function upsertFriendshipService({
           timestamp: Date.now(),
           metadata
         }),
-        catalystClient.getEntityByPointer(parsedRequest.user)
+        catalystClient.getEntitiesByPointers([context.address, parsedRequest.user!])
```
aleortega, line 99: Did you confirm that the Catalyst do always return profiles in ordered based on the elements in the received array? 🤔 

Besides, should we ensure that the profiles are available in the Catalyst before sending the update to the channel?

aleortega (approved):

aleortega (approved):

PR #56, opened 2025-02-07
feat: RPC context component

kevinszuchet:


src/adapters/rpc-server/rpc-server.ts
```diff
@@ -79,38 +74,31 @@ export async function createRpcServerComponent({
...
     },
     attachUser({ transport, address }) {
       const normalizedAddress = normalizeAddress(address)
 
       transport.on('close', () => {
-        if (SHARED_CONTEXT.subscribers[normalizedAddress]) {
-          delete SHARED_CONTEXT.subscribers[normalizedAddress]
-          logger.debug('User disconnected and removed from subscribers', { address: normalizedAddress })
-        }
+        subscribersContext.removeSubscriber(normalizedAddress)
```
aleortega, line 87: What about moving the normalized address within the subscribers ' context so clients can abstract of doing that? Or do we also need them normalized for other workflows here?

src/logic/updates.ts
```diff
@@ -39,31 +44,28 @@ function handleUpdate<T extends keyof SubscriptionEventsEmitter>(handler: Update
...
-
-    const onlineSubscribers = Object.keys(context.subscribers)
+    const onlineSubscribers = rpcContext.getSubscribersAddresses()
     const friends = await db.getOnlineFriends(update.address, onlineSubscribers)
 
     friends.forEach(({ address: friendAddress }) => {
       const normalizedAddress = normalizeAddress(friendAddress)
 
-      const emitter = context.subscribers[normalizedAddress]
+      const emitter = rpcContext.getSubscriber(normalizedAddress)
```
aleortega, line 68: Also maybe we can add it on rpcContext if it also needed there (still talking about address normalization)

src/logic/updates.ts
```diff
@@ -81,7 +83,7 @@ export async function* handleSubscriptionUpdates<T, U>({
   parseArgs
 }: SubscriptionHandlerParams<T, U>): AsyncGenerator<T> {
   const normalizedAddress = normalizeAddress(rpcContext.address)
-  const eventEmitter = rpcContext.subscribers[normalizedAddress]
+  const eventEmitter = rpcContext.subscribersContext.getSubscribers()[normalizedAddress]
```
aleortega, line 86: Is it worth adding a `subscribersContext.getSubscriber(address: string)`?

aleortega: Oh it seems we already have a `getSubscriber` right? can we use it here?

aleortega (approved):

PR #57, opened 2025-02-07
refactor: Use Lambdas client for fetching profiles

kevinszuchet:
- [x] Remove PROFILE_IMAGES_URL from defintions


src/logic/notifications.ts
```diff
@@ -1,17 +1,17 @@
...
 
 type NotificationContext = {
   requestId: string
   senderAddress: string
   receiverAddress: string
-  senderProfile: Entity
-  receiverProfile: Entity
-  profileImagesUrl: string
+  senderProfile: Pick<Profile, 'avatars'>
+  receiverProfile: Pick<Profile, 'avatars'>
```
aleortega, line 14: This Pick will make them to be Arrays, right?

src/logic/profiles.ts
```diff
@@ -1,15 +1,40 @@
...
   return avatar
 }
 
-export function getProfilePictureUrl(baseUrl: string, { id }: Pick<Entity, 'id'>): string {
-  if (!baseUrl) throw new Error('Missing baseUrl for profile picture')
+export function getProfileName(profile: Pick<Profile, 'avatars'>): string {
+  const { name } = getProfileAvatarItem(profile)
 
-  return `${baseUrl}/entities/${id}/face.png`
+  if (!name) throw new Error('Missing profile avatar name')
```
aleortega, line 15: So far so good, but we have an open issue regarding profiles being nameless and we still need to define what to do with them (allow them or do not allow them at Catalyst level), just a heads-up to keep the conversation going 👼

src/logic/profiles.ts
```diff
@@ -1,15 +1,40 @@
...
+
+  return name
+}
+
+export function getProfileUserId(profile: Pick<Profile, 'avatars'>): string {
```
aleortega, line 20: what about
```suggestion
export function getProfileAddress(profile: Pick<Profile, 'avatars'>): string {
```

src/logic/profiles.ts
```diff
@@ -1,15 +1,40 @@
...
+}
+
+export function getProfilePictureUrl(profile: Pick<Profile, 'avatars'>): string {
+  const { avatar } = getProfileAvatarItem(profile)
+  const { face256 } = avatar?.snapshots ?? {}
+
+  if (!face256) throw new Error('Missing profile avatar picture url')
+
+  return face256
 }
```
aleortega, line 40: Not for this PR, but I think we are trusting too much in the service that generates profile images. Can we discuss having a default profile image?

Currently, it seems that if a user (either the sender or the receiver) is missing a proper profile image, that user can't use these workflows. I'd try to change this behavior to keep `social-service` operational with a default image profile even when the image wasn't generated. At some point, it will get generated, and it will be rendered on UI, but even if it is missing, you can still use these workflows.

test/unit/adapters/catalyst-client.spec.ts
```diff
@@ -29,96 +31,103 @@ jest.mock('../../../src/utils/timer', () => ({
...
+const CATALYST_LAMBDAS_LOAD_BALANCER_URL = 'http://catalyst-server.com/lambdas'
 
 describe('Catalyst client', () => {
   let catalystClient: ICatalystClientComponent
-  let contentClientMock: ContentClient
+  let lambdasClientMock: LambdasClient
 
   beforeEach(async () => {
-    mockConfig.requireString.mockResolvedValue(CATALYST_CONTENT_LOAD_BALANCER_URL)
+    mockConfig.requireString.mockResolvedValue(CATALYST_LAMBDAS_LOAD_BALANCER_URL)
```
aleortega, line 41: Just a heads-up, this is working because we are mocking directly `requireString` but I think I saw that the env var name wasn't change on the real implementation.

For future config mocks maybe it is worth placing them at `test/components.ts` directly while overlapping the initComponents.config so you can detect this kind of missmatch in tests 👼

aleortega (approved):
I left some comments and NITs to discuss them in the future but besides that this PR looks good to me, so approving it to prevent blocking :)

PR #69, opened 2025-02-25
feat: Add block users feature

kevinszuchet:


src/adapters/rpc-server/services/unblock-user.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export function unblockUserService({
+  components: { logs, db, catalystClient, pubsub }
+}: RPCServiceContext<'logs' | 'db' | 'catalystClient' | 'pubsub'>) {
+  const logger = logs.getLogger('unblock-user-service')
+
+  return async function (request: UnblockUserPayload, context: RpcServerContext): Promise<UnblockUserResponse> {
+    try {
+      const { address: blockerAddress } = context
+      const blockedAddress = request.user?.address
```
aleortega, line 17: Same here, no need to validate address?

src/adapters/rpc-server/services/unblock-user.ts
```diff
@@ -0,0 +1,88 @@
...
+          }
+        }
+      }
+
+      const profile = await catalystClient.getProfile(blockedAddress)
+
+      if (!profile) {
+        return {
+          response: {
+            $case: 'internalServerError',
```
aleortega, line 33: Could we iterate over errors being returned in the future? so we prevent returning internalServerError on not founds?

src/logic/blocks.ts
```diff
@@ -0,0 +1,39 @@
...
+import { BlockUserWithDate, SubscriptionEventsEmitter } from '../types'
+
+export function parseProfileToBlockedUser(profile: Profile, blockedAt?: Date): BlockedUserProfile {
+  const { name, userId, hasClaimedName, profilePictureUrl } = getProfileInfo(profile)
+
+  return {
+    address: userId,
+    name,
+    hasClaimedName,
+    profilePictureUrl,
```
aleortega, line 16: Are we going to show real profile picture of blocked user?

src/migrations/1737745803297_add-blocks.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    blocker_address: {
+      type: PgType.VARCHAR,
+      notNull: true
+    },
+    blocked_address: {
+      type: PgType.VARCHAR,
+      notNull: true
+    },
+    blocked_at: {
```
aleortega, line 19: do we need to sort/paginate blocked users as well? shall we add a new index for blocked_at to make that sort/pagination faster or do you think it is not needed?

src/migrations/1737745803297_add-blocks.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+  })
+
+  pgm.createIndex('blocks', ['blocker_address'])
+  pgm.createIndex('blocks', ['blocked_address'])
+  pgm.createIndex('blocks', ['blocker_address', 'blocked_address'], { unique: true })
+}
+
+export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.dropTable('blocks')
```
aleortega, line 31: I think you first need to drop relations (indexes) and then execute this dropTable 👼

aleortega (approved):

src/adapters/rpc-server/services/unblock-user.ts
kevinszuchet: I had it in my backlog and kept postponing it. I will do it now.

PR #75, opened 2025-03-19
feat: Enhance Metrics

kevinszuchet:


LautaroPetaccio (approved):
Looks great! We'll have the same quality of metrics like we have for the HTTP servers with this implementation. Great job 🥇

PR #81, opened 2025-04-04
fix: Add authenticating flag and reject messages while user is authenticating

kevinszuchet:


src/controllers/handlers/ws-handler.ts
```diff
@@ -26,19 +27,148 @@ export async function registerWsHandler(
...
+      const transport = await createUWebSocketTransport(ws, eventEmitter, config, logs)
+
+      changeStage(data, {
+        auth: true,
+        address,
+        eventEmitter,
+        isConnected: true,
+        transport,
+        authenticating: false
+      })
```
LautaroPetaccio, line 101: Although this state change will work perfectly, what if we move it down as the last thing to do in the scope so no-one is tempted to put an async call after it which will switch to run something else concurrently and leave the `rpcServer` calls and callbacks without being set properly?

LautaroPetaccio (approved):
Looks great 🚀 , good catch!

src/controllers/handlers/ws-handler.ts
kevinszuchet: Moved it to the `finally` clause. Good catch!

PR #91, opened 2025-04-22
fix: Enhance backpressure and drain handling

kevinszuchet:


src/utils/UWebSocketTransport.ts
```diff
@@ -80,29 +82,41 @@ export async function createUWebSocketTransport<T extends { isConnected: boolean
...
         } catch (error: any) {
-          item.future.reject(new Error(`Failed to send message: ${error.message}`))
+          const errorMessage = `Failed to send message: ${error.message}`
+          item.future.reject(new Error(errorMessage))
           messageQueue.shift()
-          cleanup()
+          events.emit('error', new Error(errorMessage))
           return
         }
         await new Promise((resolve) => setTimeout(resolve, 0))
```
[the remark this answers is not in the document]
kevinszuchet: You make a great point.

The reason behind this approach is to prevent the event loop from blocking during message processing by periodically yielding control. Although your concern about race conditions is valid, we have implemented some safeguards to address these issues. 

However, we can enhance this process by:
1. Using `setImmediate` instead of `setTimeout`, which seems more appropriate.
2. Processing messages in batches to minimize yields while still maintaining non-blocking behavior.

WDYT?

src/utils/UWebSocketTransport.ts
```diff
@@ -80,29 +82,41 @@ export async function createUWebSocketTransport<T extends { isConnected: boolean
...
+            // Send failed, socket might be closed or backpressured
+            if (!isTransportActive || !socket.getUserData().isConnected) {
+              // If transport is not active or socket is closed, reject with connection closed
+              const error = new Error('Connection closed')
+              item.future.reject(error)
+              messageQueue.shift()
+              events.emit('error', error)
+              return
+            } else {
+              // If transport is active but send failed, retry later
```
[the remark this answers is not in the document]
kevinszuchet: Great catch 🙏

src/utils/UWebSocketTransport.ts
```diff
@@ -35,80 +43,159 @@ export async function createUWebSocketTransport<T extends { isConnected: boolean
...
-          item.future.reject(new Error(`Failed to send message: ${error.message}`))
-          messageQueue.shift()
-          cleanup()
-          return
+          continue
+        }
+
+        const result = processNextMessage(currentMessage)
+
+        if (result === UWebSocketSendResult.SUCCESS || result === UWebSocketSendResult.ERROR) {
```
LautaroPetaccio, line 105: Whenever the [backpressure status code is returned](https://github.com/uNetworking/uWebSockets/blob/master/misc/READMORE.md#backpressure), it means that the message fits in the backpressure buffer and will be queued to be sent whenever it's possible. So that means that we shouldn't retry sending the message to the user.
We should only retry it when there's a DROPPED result, which means that the backpressure is too high to be a able to send a message.
```suggestion
        if (result === UWebSocketSendResult.SUCCESS || result === UWebSocketSendResult.ERROR || result === UWebSocketSendResult.ERROR) {
```

src/utils/UWebSocketTransport.ts
```diff
@@ -35,80 +43,159 @@ export async function createUWebSocketTransport<T extends { isConnected: boolean
...
+
+      switch (result) {
+        case UWebSocketSendResult.SUCCESS:
+          item.future.resolve()
+          messageQueue.shift()
+          break
+
+        case UWebSocketSendResult.BACKPRESSURE:
+          metrics.increment('ws_backpressure_events', { result: 'backpressure' })
+          item.attempts++ // Increment attempts to enable exponential backoff
```
LautaroPetaccio, line 159: Following my previous message, these type of messages will be queued to be sent whenever possible, so we should move the retry process to the dropped ones if we want to give them a change to be sent, WDYT?

src/utils/UWebSocketTransport.ts
```diff
@@ -35,80 +43,159 @@ export async function createUWebSocketTransport<T extends { isConnected: boolean
...
+
+        case UWebSocketSendResult.DROPPED:
+          // Message was dropped by the underlying library due to maxBackpressure limit.
+          // Do not retry this specific message, reject its future and remove from queue.
+          logger.warn('Message dropped due to backpressure limit', {
+            transportId,
+            messageSize: item.message.byteLength
+          })
+          metrics.increment('ws_backpressure_events', { result: 'dropped' })
+          item.future.reject(new Error('Message dropped due to backpressure limit'))
```
LautaroPetaccio, line 170: The dropped messages should be either rejected or the user disconnected if we've reached a maximum amount of retries. We might want to go with the disconnection logic, as if we drop this message and newer messages eventually arrive due to the backpressure being freed, the client might be kept waiting for a response that could never arrive, right?

LautaroPetaccio (commented):
Great job! 🎖️ , I've left a couple of comments.

src/utils/UWebSocketTransport.ts
mikhail-dcl: Is there anything the client can do if it receives this error?

src/utils/UWebSocketTransport.ts
```diff
@@ -35,80 +43,159 @@ export async function createUWebSocketTransport<T extends { isConnected: boolean
...
-  const messageQueue: Array<{ message: Uint8Array; future: IFuture<void> }> = []
+  const retryDelayMs = (await config.getNumber('WS_TRANSPORT_RETRY_DELAY_MS')) || 1000
+  const maxRetryAttempts = (await config.getNumber('WS_TRANSPORT_MAX_RETRY_ATTEMPTS')) || 5
+  const maxBackoffDelayMs = (await config.getNumber('WS_TRANSPORT_MAX_BACKOFF_DELAY_MS')) || 30000 // 30 seconds maximum delay
 
   let isTransportActive = true
   let isInitialized = false
+
+  // Simple queue for messages that couldn't be sent immediately
+  const messageQueue: Array<{
```
LautaroPetaccio, line 61: Should we keep the message queue even if we have the backpressure buffer? That is, should we keep trying to send messages after the backpressure buffer got as big as we consider it to be it's maximum size?
We can leave it for the time being, but it's something to consider.

src/utils/UWebSocketTransport.ts
```diff
@@ -35,80 +43,159 @@ export async function createUWebSocketTransport<T extends { isConnected: boolean
...
+          // We should retry this message with exponential backoff
+          logger.warn('Message dropped due to backpressure limit, will retry', {
+            transportId,
+            messageSize: item.message.byteLength,
+            attempts: item.attempts
+          })
+          metrics.increment('ws_backpressure_events', { result: 'dropped' })
+          item.attempts++ // Increment attempts to enable exponential backoff
+          break
+
```
LautaroPetaccio, line 175: Should we measure any uWS error statuses as well?

test/unit/utils/UWebSocketTransport.spec.ts
```diff
@@ -10,270 +15,693 @@ describe('UWebSocketTransport', () => {
...
-      expect(transport.isConnected).toBe(true)
+  describe('Transport Lifecycle', () => {
+    it('should initialize with correct state and emit connect event', async () => {
+      // Set up connect listener before creating transport
+      const connectListener = jest.fn()
+      const newEmitter = mitt<IUWebSocketEventMap>()
+
+      // Create new transport instance
+      const newTransport = await createUWebSocketTransport(mockSocket, newEmitter, mockConfig, mockLogs, mockMetrics)
+      newTransport.on('connect', connectListener) // Add listener for completeness, but timing makes it hard to test reliably here
```
LautaroPetaccio, line 59: Isn't the connect emitted before the transport is set? I wouldn't test it if that's the case, we can leave it out.

test/unit/utils/UWebSocketTransport.spec.ts
```diff
@@ -10,270 +15,693 @@ describe('UWebSocketTransport', () => {
...
     })
 
-    it('should handle cleanup during active message processing', async () => {
-      jest.useFakeTimers()
-      mockSocket.send.mockReturnValue(0) // Force send failure
+    afterEach(() => {
+      jest.useRealTimers()
+    })
 
+    // Coverage for lines 200-208
```
LautaroPetaccio, line 600: These comments might get old if we change some lines of code, what do you think about removing it?

test/unit/utils/UWebSocketTransport.spec.ts
```diff
@@ -10,270 +15,693 @@ describe('UWebSocketTransport', () => {
...
+          error: 'Failed to send message: Network error'
+        }
+      )
+      // Check event emitter call
+      expect(errorListener).toHaveBeenCalledWith(new Error('Failed to send message: Network error'))
+    })
 
+    it('should handle cleanup during active message processing', async () => {
       const message = new Uint8Array([1])
+      mockSocket.send.mockReturnValue(UWebSocketSendResult.DROPPED)
```
LautaroPetaccio, line 684: If I'm not wrong, isn't this testing the same as in [the lines 565](https://github.com/decentraland/social-service-ea/pull/91/files#diff-b567943d48312e33460faf772f889ffb94a09c02003ba44388e7ea94d9bd02a6R677)?

test/unit/utils/UWebSocketTransport.spec.ts
```diff
@@ -10,270 +15,693 @@ describe('UWebSocketTransport', () => {
...
+      // Ensure no more processing occurs
+      await jest.advanceTimersByTimeAsync(5000)
+      expect(mockSocket.send).toHaveBeenCalledTimes(1)
     })
   })
 
-  describe('Error Scenarios', () => {
-    it('should handle socket errors during queue processing', async () => {
+  describe('Transport Error Handling', () => {
+    it('should handle errors during message processing (send error)', async () => {
```
LautaroPetaccio, line 652: Isn't his being already tested on [the lines 119](https://github.com/decentraland/social-service-ea/pull/91/files#diff-b567943d48312e33460faf772f889ffb94a09c02003ba44388e7ea94d9bd02a6R385) and below?

test/unit/utils/UWebSocketTransport.spec.ts
```diff
@@ -10,270 +15,693 @@ describe('UWebSocketTransport', () => {
...
+          isInitialized: 'true' // Should be true at the time of logging
+        }
+      )
+    })
+  })
 
-      const sendPromise = transport.sendMessage(message)
-      jest.advanceTimersByTime(5000)
+  describe('Transport State Management', () => {
+    it('should handle transport state transitions', async () => {
```
LautaroPetaccio, line 513: Can this happen in the current flow of execution? If the user gets disconnected, doesn't the transport gets re-built?

test/unit/utils/UWebSocketTransport.spec.ts
```diff
@@ -10,270 +15,693 @@ describe('UWebSocketTransport', () => {
...
       transport.sendMessage('invalid' as any)
+
       expect(errorListener).toHaveBeenCalledWith(
         new Error('WebSocketTransport: Received unknown type of message, expecting Uint8Array')
       )
     })
+
+    it('should handle cleanup errors gracefully', () => {
+      const error = new Error('Cleanup error')
+      mockSocket.end.mockImplementation(() => {
```
LautaroPetaccio, line 425: If I'm not wrong, when sending data, the code is not using end, should we test this?

LautaroPetaccio (commented):
Great work! The code looks great!
I've left a couple of comments to review

PR #93, opened 2025-05-23
feat: Initial migrations for Communities schemas

kevinszuchet:


src/migrations/1747997882050_add-communities.ts
```diff
@@ -0,0 +1,57 @@
...
+    },
+    thumbnail_url: {
+      type: PgType.VARCHAR,
+      notNull: false
+    },
+    places: {
+      type: 'varchar[]',
+      notNull: true,
+      default: '{}'
+    },
```
LautaroPetaccio, line 33: If I remember correctly, places will be a list of places, associated with the community, that can be added, removed and probably sorted, paginated and  such. To allow us to perform those tasks, what do you think about eventually (whenever needed) a new table that will be linked to the collection and let us perform those operations?

src/migrations/1747997984897_add-community-roles.ts
```diff
@@ -0,0 +1,46 @@
...
+      default: pgm.func('now()')
+    }
+  })
+
+  // Insert default roles
+  pgm.sql(`
+    INSERT INTO community_roles (id, name, permissions) VALUES
+    (1, 'owner', '["edit_info", "add_remove_places", "accept_reject_requests", "ban_players", "send_invitations", "edit_settings", "delete_community", "assign_roles"]'),
+    (2, 'moderator', '["edit_info", "add_remove_places", "accept_reject_requests", "ban_players", "send_invitations"]'),
+    (3, 'member', '[]')
```
LautaroPetaccio, line 40: Will community roles be created, updated or deleted programmatically? Like, will an admin be able to add a new role and define the permissions? If that's not the case, what do you think about having a fixed set of roles (fixed roles). That would prevent you from requesting to the DB what the user can do every time you need to perform some action. The role will be defined as a string for each community member.

src/migrations/1747998541677_add-community-members.ts
```diff
@@ -0,0 +1,60 @@
...
+      notNull: false
+    },
+    kicked_at: {
+      type: PgType.TIMESTAMP,
+      notNull: false
+    },
+    kick_reason: {
+      type: PgType.TEXT,
+      notNull: false
+    }
```
LautaroPetaccio, line 45: This table seems to hold the information about a member of a community. If a community member is kicked, we might want to remove this entry from the table as it does not belong to the community anymore. If we choose not, to see if a user belongs to a community we would need to check for the kicked property, and whenever the user re-joins, we would need to update it properly.
If we want to keep the historic information about a user being kicked, what do you think about having a new table with that information?

src/migrations/1747998541677_add-community-members.ts
```diff
@@ -0,0 +1,60 @@
...
+  })
+
+  pgm.createIndex('community_members', 'community_id')
+  pgm.createIndex('community_members', 'member_address')
+  pgm.createIndex('community_members', ['community_id', 'member_address'], { unique: true })
+  pgm.createIndex('community_members', ['community_id', { name: 'role_id', sort: 'ASC' }])
```
LautaroPetaccio, line 51: I would hold the creation of the indexes until we have the queries to see if they fit them, what do you think about it?

src/migrations/1747998545905_add-community-bans.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    reason: {
+      type: PgType.TEXT,
+      notNull: false
+    },
+    active: {
+      type: PgType.BOOLEAN,
+      notNull: true,
+      default: true
+    }
```
LautaroPetaccio, line 40: Will we be required to store an historical list of bans? If that's not the case, what do you think about not having the `active` column? If a user gets unbanned we can just remove the entry.

LautaroPetaccio (commented):
Great job! 💯  I've left some comments to review

src/migrations/1747997882050_add-communities.ts
```diff
@@ -0,0 +1,49 @@
...
+    description: {
+      type: PgType.TEXT,
+      notNull: true
+    },
+    owner_address: {
+      type: PgType.VARCHAR,
+      notNull: true
+    },
+    thumbnail_url: {
```
aleortega, line 25: we will have many thumbnails, right? I'm thinking about these options:
* make this an array (don't like it tbh)
* rename this to be `thumbnail_id` -- so then we can build static urls like: https://cdn.org/communities/thumbnails/<id>/256x256 or https://cdn.org/communities/thumbnails/<id>/512x512 and whatever sizes we do need there
* since we already have `id` we could use that to build static URLs and do not have any reference to thumbnails in the DB

aleortega (commented):
LGTM!

yet, i left some comments about some doubts I do have 👼

aleortega (approved):
LGTM!

PR #94, opened 2025-05-26
feat: Get and Delete Communities

kevinszuchet:


src/adapters/communities-db.ts
```diff
@@ -7,5 +19,123 @@ export function createCommunitiesDBComponent(
...
+            ELSE world_name
+          END as place
+        FROM community_places
+        WHERE community_id = ${communityId}`
+      const result = await pg.query<{ place: string }>(query)
+      return result.rows.map((row) => row.place)
+    },
+
+    async getCommunities(memberAddress: string, options?: { pagination?: Pagination }) {
+      const { pagination = { limit: 10, offset: 0 } } = options ?? {}
```
aleortega, line 80: should we set a default global limit to use in all queries that requires pagination?

src/adapters/communities-db.ts
```diff
@@ -7,5 +19,123 @@ export function createCommunitiesDBComponent(
 
   const logger = logs.getLogger('communities-db-component')
 
-  return {}
+  return {
+    async getCommunity(id: string, userAddress: string) {
```
aleortega, line 23: can we add returning types?

src/adapters/errors.ts
```diff
@@ -0,0 +1,8 @@
+import { NotFoundError } from '@dcl/platform-server-commons'
+
+export class CommunityNotFoundError extends NotFoundError {
```
aleortega, line 3: not a strong opinion but i'd try to avoid managing flows through exceptions.

Maybe we can have this error to be a HTTP returning type that we can re-use in our handlers when we detect that the requesting community does not exists?

src/controllers/handlers/delete-community-handler.ts
```diff
@@ -0,0 +1,57 @@
...
+): Promise<HTTPResponse> {
+  const {
+    components: { communitiesDb, logs },
+    params: { id },
+    verification
+  } = context
+  const logger = logs.getLogger('privacy-handler')
+  logger.info(`Deleting community: ${id}`)
+
+  const userAddress = verification?.auth.toLowerCase()
```
aleortega, line 21: if singed fetch is required on this endpoint we will always have a verification.auth so we can do:

`verification!.auth` and then only validate the address with `EthAddress.validate`

src/controllers/handlers/delete-community-handler.ts
```diff
@@ -0,0 +1,57 @@
...
+    HandlerContextWithPath<'communitiesDb' | 'logs', '/communities/:id'>,
+    'components' | 'params' | 'verification'
+  >
+): Promise<HTTPResponse> {
+  const {
+    components: { communitiesDb, logs },
+    params: { id },
+    verification
+  } = context
+  const logger = logs.getLogger('privacy-handler')
```
aleortega, line 18: rename logger

src/controllers/handlers/get-community-handler.ts
```diff
@@ -0,0 +1,53 @@
...
+    'url' | 'components' | 'params' | 'verification'
+  >
+): Promise<HTTPResponse<CommunityWithMembersCount>> {
+  const {
+    components: { communitiesDb, logs },
+    params: { id },
+    verification
+  } = context
+  const userAddress = verification?.auth.toLowerCase()
+  const logger = logs.getLogger('privacy-handler')
```
aleortega, line 19: rename also this logger and maybe also check others 🙏

src/logic/community-roles.ts
```diff
@@ -1,48 +1,35 @@
...
-  | 'send_invitations'
-  | 'edit_settings'
-  | 'delete_community'
-  | 'assign_roles'
+import { CommunityRole, CommunityPermission } from '../types/entities'
 
-export type CommunityRole = 'owner' | 'moderator' | 'member'
+export const OWNER_PERMISSIONS: CommunityPermission[] = [
+  'edit_info',
+  'add_remove_places',
```
aleortega, line 5: what do you think about splitting these grouped permissions in two?
`add_remove_places` and `accept_reject_requests` just to be more granular and prevent migrations in the future if a new role able to accept request but not to reject them appears in the wild

src/types/entities.ts
```diff
@@ -66,3 +66,50 @@ export type Pagination = {
...
+  privacy: 'public' | 'private'
+  active: boolean
+}
+
+export type CommunityWithMembersCount = Community & {
+  membersCount: number
+}
+
+export type CommunityPlace = {
+  position: string
```
aleortega, line 113: both of them should be nullables, right?

aleortega (approved):

src/adapters/communities-db.ts
kevinszuchet: Isn't it enough with the `ICommunitiesDatabaseComponent` type?

src/adapters/errors.ts
kevinszuchet: I considered creating a new exception, knowing we use the error handler that catches NotFoundError to return the proper 404 response.

PR #95, opened 2025-05-28
feat: Start new HTTP Server + DB renamed to Friends DB

kevinszuchet:
- New HTTP Server listening in HTTP_SERVER_PORT
- UWS HTTP Server + WS Server now listening in UWS_HTTP_SERVER_PORT
- Refactor: DB Component now is Friends DB Component
- Split types file for better readability


src/types/http.ts
```diff
@@ -0,0 +1,41 @@
...
+  headers?: Record<string, string>
+  body?: ResponseBody
+}
+
+export type IHandler = {
+  path: string
+  f: (res: HttpResponse, req: HttpRequest) => Promise<IHandlerResult>
+}
+
+export type Pagination = {
```
aleortega, line 29: NIT: this more like internal Pagination for DB, right?

HTTP Pagination should have pageSize and pageNum according to what we discussed.

PS: we can move this later, just a heads-up 👼

aleortega (approved):
❤️

PR #97, opened 2025-05-28
feat: Add the start private voice chat request

LautaroPetaccio:
This PR adds the start call request RPC call, which indicates the intent to start a call with someone.
In this PR, the `settings` and the `voice` logic components are created, containing the business logic for the usage of those components and orchestrating the use of the adapter components.
The use `settings` component is not fully migrated to its logical component. There's still a `utils` file that holds most of the logic that is still in use.

Some information about voice chats:
- When making a private voice chat, if the permissions are correct, a DB entry with the voice chat information and an expiration time will be created.
- Upon accepting or rejecting the voice chat, the voice chat entry will be deleted.
- The could be more than one voice chats in the DB if there are some that are expired.
- If one of the users gets disconnected, having requested the intent to initiate a voice call before, the call will be rejected and deleted. TBD but that's the idea.


src/adapters/voice-db.ts
```diff
@@ -0,0 +1,35 @@
...
+    async areUsersBeingCalledOrCallingSomeone(userAddresses: string[]): Promise<boolean> {
+      const normalizedUserAddresses = userAddresses.map(normalizeAddress)
+
+      const query = SQL`
+        SELECT EXISTS (
+          SELECT 1 FROM private_voice_chats WHERE caller_address IN (${normalizedUserAddresses}) OR callee_address IN (${normalizedUserAddresses})
+        )
+      `
+      const results = await pg.query<{ exists: boolean }>(query)
+      return results.rows[0].exists
```
kevinszuchet, line 23: I added a method `exists` in a decorated `pg` component that you could use here if you want.

src/logic/settings/utils.ts
```diff
@@ -4,11 +4,12 @@ import {
   BlockedUsersMessagesVisibilitySetting as RPCBlockedUsersMessagesVisibilitySetting
 } from '@dcl/protocol/out-js/decentraland/social_service/v2/social_service_v2.gen'
 import {
+  AppComponents,
```
kevinszuchet, line 7:
```suggestion
```

test/unit/logic/voice-logic.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+  createPrivateVoiceChatMock = jest.fn()
+  const logs: ILoggerComponent = {
+    getLogger: () => ({
+      info: () => undefined,
+      error: () => undefined,
+      debug: () => undefined,
+      warn: () => undefined,
+      log: () => undefined
+    })
+  }
```
kevinszuchet, line 48: Should we use the mock?

kevinszuchet (approved):
LGTM! Just left some tiny comments

PR #100, opened 2025-05-30
feat: Get Public Communities

kevinszuchet:


src/adapters/pg.ts
```diff
@@ -0,0 +1,42 @@
+import { IBaseComponent } from '@well-known-components/interfaces'
+import { createPgComponent as createBasePgComponent, Options } from '@well-known-components/pg-component'
+import { PoolClient } from 'pg'
+import { IPgComponent } from '../types'
+import { SQLStatement } from 'sql-template-strings'
+
+export async function createPgComponent(
```
aleortega, line 7: greattt, thx!

src/logic/community/community.ts
```diff
@@ -39,6 +45,19 @@ export function createCommunityComponent(
       }
     },
 
+    getPublicCommunities: async (options: GetCommunitiesOptions): Promise<GetCommunitiesResult<PublicCommunity>> => {
```
aleortega, line 48: What do you think about, maybe another one, but seems we will have the `public | private` concept it may confuses future devs, not strong opinion btw
```suggestion
    getCommunitiesWithPublicInformation: async (options: GetCommunitiesOptions): Promise<GetCommunitiesResult<PublicCommunity>> => {
```

src/logic/community/types.ts
```diff
@@ -3,7 +3,11 @@ import { CommunityRole, Pagination } from '../../types/entities'
 
 export type ICommunityComponent = {
   getCommunity: (id: string, userAddress: string) => Promise<CommunityWithMembersCount>
-  getCommunities: (userAddress: string, options: GetCommunitiesOptions) => Promise<GetCommunitiesResult>
+  getCommunities: (
+    userAddress: string,
+    options: GetCommunitiesOptions
+  ) => Promise<GetCommunitiesResult<CommunityResult>>
```
aleortega, line 9: Super NIT: Result sounds like an HTTP response but here we are declaring the return from the logic component which should be HTTP agnostic 🤔 

```suggestion
  ) => Promise<GetCommunitiesResult<CommunityResult>>
```

aleortega (approved):
LGTM!

PR #121, opened 2025-06-04
feat: Include friendship status when getting members

kevinszuchet:


aleortega (approved):
LGTM, I only left two comments regarding Omitting from another type

PR #123, opened 2025-06-05
feat: Improve actions to prevent race conditions

LautaroPetaccio:
This PR adds functions that prevent race conditions by checking if the deletion of the private voice chat (which is done after the intent of a private voice chat is accepted or rejected) occurred. 
If it didn't occurred while:
- Accepting a private voice chat: It means it was rejected, meaning that we need to roll back accepting a voice chat.
- Rejecting a private voice chat: It means that it was already rejected (possible a doble reject or a cancelled call) and we need to not do anything to the call and notify the user executing the function about it.

### DIFF
--- a/test/unit/adapters/commms-gatekeepr.spec.ts
+++ b/test/unit/adapters/commms-gatekeepr.spec.ts
@@ -1,7 +1,7 @@
-import { ILoggerComponent, IFetchComponent } from '@well-known-components/interfaces'
+import { ILoggerComponent, IFetchComponent, IConfigComponent } from '@well-known-components/interfaces'
 import { createCommsGatekeeperComponent } from '../../../src/adapters/comms-gatekeeper'
 import { ICommsGatekeeperComponent, PrivateMessagesPrivacy } from '../../../src/types'
-import { mockConfig } from '../../mocks/components'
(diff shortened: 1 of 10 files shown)


test/unit/adapters/commms-gatekeepr.spec.ts
```diff
@@ -1,7 +1,7 @@
-import { ILoggerComponent, IFetchComponent } from '@well-known-components/interfaces'
+import { ILoggerComponent, IFetchComponent, IConfigComponent } from '@well-known-components/interfaces'
```
kevinszuchet, line 1: Would you mind renaming the file to comms-gatekeeper.spec.ts 😋?

test/unit/adapters/commms-gatekeepr.spec.ts
```diff
@@ -207,3 +216,72 @@ describe('when getting the private voice chat credentials', () => {
...
+        ok: true,
+        status: 200
+      })
+    })
+
+    it('should resolve successfully', async () => {
+      await expect(commsGatekeeper.endPrivateVoiceChat(callId, address)).resolves.toBeUndefined()
+    })
+
+    it('should make the correct API call', async () => {
```
kevinszuchet, line 241: Should we use a more specific description like "should make a delete request to [...]" or similar?

PR #125, opened 2025-06-06
feat: validate names ownership while creating a community

aleortega:


src/adapters/catalyst-client.ts
```diff
@@ -59,5 +60,24 @@ export async function createCatalystClient({
...
   }
 
-  return { getProfiles, getProfile }
+  async function getOwnedNames(
+    address: EthAddress,
+    options: ICatalystClientRequestOptions = {}
+  ): Promise<OwnedNames[]> {
+    const { retries = 3, waitTime = 300, lambdasServerUrl } = options
+    const executeClientRequest = rotateLambdasServerClient(
+      (lambdasClientToUse) => lambdasClientToUse.getNames(address),
```
kevinszuchet, line 69: Should we use a small page size here since we only want to know if the user owns at least one name?

Maybe we can accept a new parameter of type `GetNamesParams` in the `getOwnedNames`, so the logic component can use a page size of one. WDYT?

kevinszuchet (approved):
LGTM!

src/types/entities.ts
```diff
@@ -95,3 +95,10 @@ export enum CommunityRole {
   Member = 'member',
   None = 'none'
 }
+
+export type OwnedNames = {
```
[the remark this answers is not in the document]
aleortega: truee

src/adapters/catalyst-client.ts
aleortega: will do, great catch

PR #126, opened 2025-06-06
feat: Get incoming private voice chat requests

### LINKED ISSUE #132: Add the get incoming private voice chat requests
As we can't know if a another user is connected from the context of a user in the social service, the client requires to ask for the incoming private voice chat requests upon connecting to the server.
In order to complete this task, create the handler for the get incoming private voice chat request that should return any incoming chat requests.

LautaroPetaccio:
This PR adds the handler to get the incoming private voice chat requests, that is, the call that the user is receiving at the moment.

Closes #132

### DIFF
--- a/jest.integration.config.js
+++ b/jest.integration.config.js
@@ -1,5 +1,6 @@
 module.exports = {
   preset: 'ts-jest',
+  runInBand: true,
   transform: {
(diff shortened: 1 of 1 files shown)


kevinszuchet (approved):
LGTM!

PR #140, opened 2025-06-09
feat: Get online members from community

kevinszuchet:


src/controllers/handlers/update-member-role-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+  try {
+    const updaterAddress = verification!.auth.toLowerCase()
+    const targetAddressLower = targetAddress.toLowerCase()
+
+    if (!EthAddress.validate(targetAddressLower)) {
+      throw new InvalidRequestError(`Invalid address provided`)
+    }
+
+    const searchParams = new URLSearchParams(url.search)
+    const newRole = searchParams.get('role')
```
aleortega, line 32: Will we receive the role to set through query parameters?

src/controllers/handlers/update-member-role-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    if (!newRole || !Object.values(CommunityRole).includes(newRole as CommunityRole)) {
```
aleortega, line 34: Can we move this check within `community.updateMemberRole` so every piece of code that calls it gets this validation?

src/logic/community/community.ts
```diff
@@ -25,12 +26,50 @@ import {
...
 export function createCommunityComponent(
-  components: Pick<AppComponents, 'communitiesDb' | 'catalystClient' | 'communityRoles' | 'logs'>
+  components: Pick<AppComponents, 'communitiesDb' | 'catalystClient' | 'communityRoles' | 'logs' | 'peersStats'>
 ): ICommunityComponent {
-  const { communitiesDb, catalystClient, communityRoles, logs } = components
+  const { communitiesDb, catalystClient, communityRoles, logs, peersStats } = components
 
   const logger = logs.getLogger('community-component')
 
+  const getCommunityMembers = async (id: string, options: GetCommunityMembersOptions, userAddress?: EthAddress) => {
```
aleortega, line 35: Can we name this function differently so it does not have the same name as the exported one, `getCommunityMembers`?

src/logic/community/community.ts
```diff
@@ -25,12 +26,50 @@ import {
...
+
+    if (onlyOnline) {
+      onlinePeers = await peersStats.getConnectedPeers()
+      logger.info(`Getting community members for community using the ${onlinePeers.length} connected peers`)
+    }
+
+    const communityMembers = await communitiesDb.getCommunityMembers(id, {
+      userAddress,
+      pagination,
+      onlinePeers
```
aleortega, line 59: What about renaming this filter to something like `membersFilter` or another one that represents it is a member filter (which can be used for multiple purposes, in this case retrieving onlinePeers)

PR #142, opened 2025-06-10
docs: Deploying preview

kevinszuchet:


docs/communities-openapi.yaml
```diff
@@ -14,20 +31,55 @@ paths:
         Lists all communities.
         This endpoint has optional Signed Fetch authentication.
         When signed, it includes properties related to the authenticated user in its response.
+      security:
+        - SignedFetch: []
+      parameters:
+        - name: page
```
aleortega, line 37: This should be `offset`

PR #154, opened 2025-06-13
fix: Return the credentials url instead of token and url

LautaroPetaccio:
This PR changes the credentials returned for the voice chat so they return whole built credentials url.


braianj (approved):
🎉

PR #164, opened 2025-06-20
feat: validate places ownership on community creation

aleortega:
Added validation to ensure users can only add places they own when creating communities. This prevents unauthorized users from associating places they don't control with communities.


src/logic/community/community.ts
```diff
@@ -300,14 +300,43 @@ export async function createCommunityComponent(
...
+        memberAddress: community.ownerAddress,
+        role: CommunityRole.Owner
+      })
+
+      placeIds.length > 0 && (await communityPlaces.addPlaces(newCommunity.id, community.ownerAddress, placeIds))
+
+      logger.info('Community created', {
+        communityId: newCommunity.id,
+        name: newCommunity.name,
+        owner: community.ownerAddress.toLowerCase()
```
kevinszuchet, line 338: Maybe we could add the number of places attached to the community

kevinszuchet (approved):

PR #170, opened 2025-06-23
feat: referral notification

braianj:


test/unit/logic/referral.spec.ts
```diff
@@ -254,6 +391,117 @@ describe('referral-component', () => {
...
+              referrer: validReferrer,
+              invited_user: validInvitedUser,
+              status: ReferralProgressStatus.SIGNED_UP
+            }
+          ])
+          mockReferralDb.updateReferralProgress.mockResolvedValueOnce(undefined)
+          mockReferralDb.countAcceptedInvitesByReferrer.mockResolvedValueOnce(invitedUsers)
+        })
+
+        it(`should calculate correct tier for ${invitedUsers} invited users`, async () => {
```
LautaroPetaccio, line 458: As these tests are expecting a notification to be sent, what do you think about changing the `it` so it reflects what's expected?
```suggestion
        it(`should send the notification with the correct tier for ${invitedUsers} invited users`, async () => {
```

PR #174, opened 2025-06-25
refactor: Move rpc service creators to controllers handlers

kevinszuchet:


src/controllers/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './handlers/http'
+export * from './handlers/uws'
+// export * from './handlers/rpc'
```
LautaroPetaccio, line 3: This was left commented, but I think that it might be better if neither them exported here, to avoid any confusion between the HTTP and RPC controllers. What do you think about it?
Also, we could export the handlers for the indexes of each protocol so the imports are less verbose.

src/controllers/handlers/rpc/index.ts
LautaroPetaccio, line 1: Maybe we can use it to export all RPC handlers?

src/controllers/routes/http.routes.ts
```diff
@@ -1,27 +1,27 @@
...
-import { getBannedMembersHandler } from '../handlers/get-banned-members-handler'
-import { createReferralHandler } from '../handlers/create-referral-handler'
-import { updateReferralSignedUpHandler } from '../handlers/update-referral-signed-up-handler'
-import { getInvitedUsersAcceptedHandler } from '../handlers/get-invited-users-accepted-handler'
-import { updateMemberRoleHandler } from '../handlers/update-member-role-handler'
+import { getCommunityMembersHandler } from '../handlers/http/get-community-members-handlers'
+import { getMemberCommunitiesHandler } from '../handlers/http/get-member-communities-handlers'
+import { removeMemberFromCommunityHandler } from '../handlers/http/remove-member-from-community-handler'
+import { createCommunityHandler } from '../handlers/http/create-community-handler'
+import { addMemberToCommunityHandler } from '../handlers/http/add-member-to-community-handler'
```
LautaroPetaccio, line 12: What if we use the index in the HTTP handlers route and impor them all from there?

LautaroPetaccio (approved):
Great job! We definitely needed some re-org and this came up perfectly 🥇

PR #175, opened 2025-06-26
feat: add referral email and reward image

braianj:


src/migrations/1750883189123_store-email-reward-image.ts
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { MigrationBuilder, ColumnDefinitions } from 'node-pg-migrate'
+
+export const ReferralEmailsColumns: ColumnDefinitions = {
```
kevinszuchet, line 4: Could be rename both columns' vars to upper snake case or camel case?

test/integration/add-referral-email-controller.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+
+        it('should set referral email with email containing spaces and return 204', async () => {
+          const response = await makeAuthenticatedRequest(components)(identity, '/v1/referral-email', 'POST', {
+            email: '  test@example.com  '
+          })
+          expect(response.status).toBe(204)
+        })
+      })
+
+      describe('with invalid email format', () => {
```
kevinszuchet, line 65: WDYT about making all the `it` statements inside this `describe` a describe on their own? Like:

```
describe('with invalid email format'
  describe('when email is missing'
    it('should respond with a 400 status code'
```

src/logic/referral/errors.ts
```diff
@@ -27,3 +27,9 @@ export class SelfReferralError extends Error {
     super(`User cannot refer themselves: ${userAddress}`)
   }
 }
+
+export class ReferralEmailUpdateTooSoonError extends Error {
```
kevinszuchet, line 31: If you'd like, you can extend this error to an `InvalidRequestError`, especially if we intend to respond to the request with a 400 status code. 

In that case, you would only need to catch that error in the handler and re-throw it.

src/controllers/handlers/http/add-referral-email-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+    await referral.setReferralEmail({ referrer: userAddress, email })
+    return {
+      status: 204
+    }
+  } catch (error) {
+    const message = errorMessageOrDefault(error)
+    logger.error(`Error setting referral email for ${userAddress}: ${message}`)
+    logger.debug('Error stack', { stack: (error as any)?.stack })
+
+    if (error instanceof ReferralInvalidInputError) {
```
kevinszuchet, line 40: We could also extend this error to InvalidRequestError. WDYT?

src/controllers/handlers/http/add-referral-email-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+      throw new InvalidRequestError(error.message)
+    }
+
+    if (error instanceof InvalidRequestError) {
+      throw error
+    }
```
kevinszuchet, line 46: Here we can catch all the instances just for visibility (`ReferralInvalidInputError` and `ReferralEmailUpdateTooSoonError`)

PR #188, opened 2025-07-07
fix: update referral only when tier is not granted

braianj:


src/logic/referral/referral.ts
```diff
@@ -172,7 +172,7 @@ export async function createReferralComponent(
       const invitedUser = validateAddress(invitedUserToFinalize, 'invitedUser')
 
       const progress = await referralDb.findReferralProgress({ invitedUser })
-      if (!progress.length) {
+      if (!progress.length || progress[0].status !== ReferralProgressStatus.TIER_GRANTED) {
```
LautaroPetaccio, line 175: Would you mind adding a test for this case?

PR #189, opened 2025-07-07
feat: Remove WS pool

LautaroPetaccio:
This PR removes the WS pool, which is being used to:
- Store the total amount of connected users.
- Store which users are connected.
But, we were not using any of this information except the amount of connected users for the purpose of having metrics. These metrics can be computed as the sum of the connected users of all instances of the service, so this is no longer required.


aleortega (approved):
Looks great to me!

PR #196, opened 2025-07-14
feat: add voice chat status info to the communities endpoint

juanmahidalgo:
This PR adds the voice chat status data for the community in the /communities endpoint

### DIFF
--- a/src/logic/community/communities.ts
+++ b/src/logic/community/communities.ts
@@ -7,10 +7,10 @@ import {
   GetCommunitiesWithTotal,
   ICommunitiesComponent,
   CommunityPublicInformation,
-  CommunityWithMembersCount,
   MemberCommunity,
   Community,
-  CommunityUpdates
+  CommunityUpdates,
+  CommunityWithMembersCountAndVoiceChatStatus
 } from './types'
 import { isOwner, toCommunityWithMembersCount, toCommunityResults, toPublicCommunity } from './utils'
 import { EthAddress } from '@dcl/schemas'
@@ -49,7 +49,7 @@ export async function createCommunityComponent(
   }
 
   return {
-    getCommunity: async (id: string, userAddress: EthAddress): Promise<CommunityWithMembersCount> => {
+    getCommunity: async (id: string, userAddress: EthAddress): Promise<CommunityWithMembersCountAndVoiceChatStatus> => {
       const [community, membersCount, voiceChatStatus] = await Promise.all([
         communitiesDb.getCommunity(id, userAddress),
         communitiesDb.getCommunityMembersCount(id),


src/logic/community/communities.ts
```diff
@@ -59,7 +68,7 @@ export async function createCommunityComponent(
         }
       }
 
-      return toCommunityWithMembersCount(community, membersCount)
+      return toCommunityWithMembersCount(community, membersCount, voiceChatStatus)
```
aleortega, line 71: Do we need to extend this function? could we just extend the type to include `voiceChatStatus` properties and handle it like we did the `ownerName`? https://github.com/decentraland/social-service-ea/pull/191/files#diff-4eed535b564dded7d489f223b66aedc5356782816d9838d5ca9874b100e5f1c0R74

src/logic/community/types.ts
```diff
@@ -185,6 +185,11 @@ export type BannedMemberProfile = BannedMember & {
 export type CommunityWithMembersCount = Community & {
   role: CommunityRole
   membersCount: number
+  voiceChatStatus: {
```
aleortega, line 188: We should add these properties to `Community` or create a new type to state that it contains data regarding voice chat. Mostly because this type is named `CommunityWithMembersCount` and these new properties are not related to that

test/integration/get-community-controller.spec.ts
```diff
@@ -14,6 +14,13 @@ test('Get Community Controller', function ({ components, spyComponents }) {
     beforeEach(async () => {
       identity = await createTestIdentity()
       address = identity.realAccount.address.toLowerCase()
+      
+      // Mock the comms gatekeeper to return a default voice chat status
+      spyComponents.commsGatekeeper.getCommunityVoiceChatStatus.mockResolvedValue({
```
aleortega, line 19: Shall we create a new context (describe) to state this? such like: `when there is an active voice chat room`

test/unit/logic/communities.spec.ts
```diff
@@ -45,24 +48,32 @@ describe('Community Component', () => {
...
+      config: mockConfig,
+      commsGatekeeper: mockCommsGatekeeper
     })
   })
 
   describe('when getting a community', () => {
     const userAddress = '0x1234567890123456789012345678901234567890'
 
     describe('and the community exists', () => {
+      const mockVoiceChatStatus = {
```
aleortega, line 60: Same, should we link this context to a new `describe`?

aleortega (commented):
LGTM, left some comments to address!

test/integration/get-community-controller.spec.ts
juanmahidalgo: the thing is, all base tests will need it and we assert that the data is there here: https://github.com/decentraland/social-service-ea/pull/196/files#diff-2826030a371cada984b2af452b168ad76dd534185f274aac4b7fa0b0fc934850R72.
I added the `and the community has no active voice chat'` to test the opposite thing. 
Do you think we still need to split them? it might get too messy to read the tests, I thought this was cleaner

PR #202, opened 2025-07-17
feat: security validation email

braianj:


test/unit/logic/referral.spec.ts
```diff
@@ -798,7 +798,7 @@ describe('referral-component', () => {
           expect(mockEmail.sendEmail).toHaveBeenCalledWith(
             'marketing@decentraland.org',
             '[Action Needed] IRL Swag Referral Tier Unlocked',
-            `<p>A user has unlocked the IRL Swag Referral Tier and provided the following email for contact: ${validEmail}</p>`
+            `A user has unlocked the IRL Swag Referral Tier and provided the following email for contact: ${validEmail}`
```
aleortega, line 801: Is this change expected? won't break email format?

aleortega (approved):
LGTM! left a doubt in comments

test/unit/logic/referral.spec.ts
braianj: we are sending that tag as text in the emails instead of as a component

PR #211, opened 2025-07-22
feat: add logic to store active community voice chats and poll their status

juanmahidalgo:
This PR adds a mechanism to poll active community voice chats until they finish and sends the notification through the stream that the call ended.


LautaroPetaccio (commented):
Great work! 👍  I've left some comments to review

PR #215, opened 2025-07-23
feat: pass role to the comms-gatekeeper when creating/joining community voice chats

juanmahidalgo:


src/logic/community-voice/community-voice.ts
```diff
@@ -142,9 +147,11 @@ export async function createCommunityVoiceComponent({
       throw new CommunityVoiceChatNotFoundError(communityId)
     }
 
+    // Get the user's role in the community for both public and private communities
+    const userRole = await communitiesDb.getCommunityMemberRole(communityId, userAddress)
```
LautaroPetaccio, line 151: What do you think about moving this request after the banned check? We can do it concurrently with the `getUserProfileData` call to make it quicker.

LautaroPetaccio (approved):
🚀

PR #216, opened 2025-07-23
feat: Increase signature expiration time

LautaroPetaccio:
Due to an increase of logs related to the expiration time of the log in signature, this PR changes the expiration time to 3 minutes to be able to work with log in processes that take a bit longer.


kevinszuchet (approved):
👌

PR #217, opened 2025-07-23
fix: Transport & Metrics

LautaroPetaccio:
This PR does the following:
- Attaches the the user to the `rpcServer` after the callbacks are set
- Removes a transport on close callback to remove subscribers that was duplicated, as it's already done in the `transport.on('close')` coming from the `ws-handler`.
- Adds missing return status codes to the metrics wrapper.
- Stops logging the not found error when getting incoming private voice chats.


kevinszuchet (approved):
LGTM!

PR #220, opened 2025-07-24
feat: add query param to filter communities by ones with active voice chats

juanmahidalgo:
This PR adds support for the `onlyWithActiveVoiceChat` and returns communities with voice chats active

### DIFF
--- a/src/logic/community/communities.ts
+++ b/src/logic/community/communities.ts
@@ -60,6 +60,29 @@ export async function createCommunityComponent(
     return `${CDN_URL}${getCommunityThumbnailPath(communityId)}`
   }
 
+  /**
+   * Helper function to filter communities with active voice chat using batch API
+   * @param communities - Array of communities to filter
+   * @returns Promise<T[]> - Filtered array containing only communities with active voice chat
+   */
+  async function filterCommunitiesWithActiveVoiceChat<T extends { id: string }>(communities: T[]): Promise<T[]> {
+    if (communities.length === 0) {
+      return communities
+    }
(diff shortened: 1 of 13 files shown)


src/logic/community/communities.ts
```diff
@@ -127,11 +128,39 @@ export async function createCommunityComponent(
...
+              return { communityId: community.id, isActive: status?.isActive ?? false }
+            } catch (error) {
+              // If we can't get the status, assume it's not active
+              logger.warn(`Could not get voice chat status for community ${community.id}`, {
+                error: isErrorWithMessage(error) ? error.message : 'Unknown error'
+              })
+              return { communityId: community.id, isActive: false }
+            }
+          })
+        )
```
kevinszuchet, line 147: WDYT about moving the get voice chat status for multiple communities inside the comms gatekeeper component?

src/logic/community/communities.ts
```diff
@@ -163,9 +192,36 @@ export async function createCommunityComponent(
...
+
+        const activeVoiceChatCommunityIds = new Set(
+          voiceChatStatuses.filter((status) => status.isActive).map((status) => status.communityId)
+        )
+
+        filteredCommunities = communitiesWithThumbnailsAndOwnerNames.filter((community) =>
+          activeVoiceChatCommunityIds.has(community.id)
+        )
+      }
```
kevinszuchet, line 220: WDYT about moving the common logic of filtering those communities with active voice chat to a new function so we can reuse it in both places?

PR #221, opened 2025-07-24
fix: Skip updates if required profile is not found

LautaroPetaccio:
This PR skips updates if the `shouldRetrieveProfile` is set to true and the profile is not found. It also sets the property as false by default to prevent any issues.


kevinszuchet (approved):
Great!

PR #228, opened 2025-07-28
feat: managed communities retrieval

aleortega:
This PR updates the Communities API to be able to retrieve users' managed communities (_user is admin or moderator of the community_).

This is needed specifically for Events integration. Currently, only owners and administrators can link events to the community. The events page needs a way to retrieve communities managed by the user who is creating the event.

In order to achieve this data extraction, two things were implemented:
**Exposed new endpoint GET /communities/:address/managed restricted by an API ADMIN TOKEN**
This is necessary to retrieve managed communities of a specific user from the Events back-end, which is unable to perform a signed fetch.

**Updated GET /communities endpoint to support filtering by roles**
The filter created for the `GET /communities/:address/managed` endpoint was also integrated into this existing endpoint as an optional. This will be useful to render the Community dropdown in https://events.decentraland.org/submit more efficiently. It prevents returning all the communities a user belongs to, to gets only the ones the user manages.


test/integration/get-managed-communities-controller.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+        })
+        const body = await response.json()
+
+        expect(response.status).toBe(500)
+        expect(body).toHaveProperty('message')
+        expect(body.message).toBe('Count query failed')
+      })
+    })
+  })
+})
```
kevinszuchet, line 332: missing EOF
```suggestion
})

```

test/integration/get-communities-controller.spec.ts
```diff
@@ -364,6 +364,168 @@ test('Get Communities Controller', function ({ components, spyComponents }) {
         })
       })
 
+      describe('when filtering by roles', () => {
```
kevinszuchet, line 367: Would you mind checking the test cases that describe test scenarios in the `it` instead of the `describe`?

test/integration/get-managed-communities-controller.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+            }
+          })
+          expect(response.status).toBe(401)
+        })
+      })
+    })
+
+    describe('and the request is authenticated with admin token', () => {
+      describe('when requesting managed communities for a user with owner and moderator roles', () => {
+        it('should respond with a 200 status code and return only communities where user is owner or moderator', async () => {
```
kevinszuchet, line 119: There are some cases here, too. Would you mind also checking this test file?

src/adapters/communities-db.ts
```diff
@@ -267,9 +269,17 @@ export function createCommunitiesDBComponent(
...
+
+      if (onlyMemberOf && roles) {
+        membersJoin = SQL` JOIN community_members cm ON c.id = cm.community_id AND cm.member_address = ${normalizedMemberAddress} AND cm.role = ANY(${roles})`
+      } else if (onlyMemberOf) {
+        membersJoin = SQL` JOIN community_members cm ON c.id = cm.community_id AND cm.member_address = ${normalizedMemberAddress}`
+      } else if (roles) {
+        membersJoin = SQL` JOIN community_members cm ON c.id = cm.community_id AND cm.member_address = ${normalizedMemberAddress} AND cm.role = ANY(${roles})`
+      } else {
+        membersJoin = SQL` LEFT JOIN community_members cm ON c.id = cm.community_id AND cm.member_address = ${normalizedMemberAddress}`
+      }
```
kevinszuchet, line 282: Do you think we could move this to the `queries.ts` file and possibly reduce the number of' if' statements in the new function so we write the least SQL possible?

I imagine something like:

```js
const baseJoin = SQL` JOIN community_members cm ON c.id = cm.community_id AND cm.member_address = ${normalizedMemberAddress}`

if (roles && roles.length > 0) {
  return baseJoin.append(SQL` AND cm.role = ANY(${roles})`)
}

if (onlyMemberOf) {
  return baseJoin
}

return SQL` LEFT `.append(baseJoin)
```

We would be able then to reuse it for the `getCommunitiesCount`.

src/controllers/handlers/http/get-communities-handler.ts
```diff
@@ -28,10 +28,20 @@ export async function getCommunitiesHandler(
   const search = url.searchParams.get('search')
   const onlyMemberOf = url.searchParams.get('onlyMemberOf')?.toLowerCase() === 'true'
   const onlyWithActiveVoiceChat = url.searchParams.get('onlyWithActiveVoiceChat')?.toLowerCase() === 'true'
+  const roles: CommunityRole[] | undefined = url.searchParams
+    .getAll('roles')
+    ?.map((role) => role as CommunityRole)
```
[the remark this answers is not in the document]
aleortega: updated the parsing to ignore invalid values!

PR #232, opened 2025-07-31
Feat/community voice chat position

juanmahidalgo:
This PRs adds:
- array of coordinates of the community places.
- community name
- community thumbnails

this will be needed in the client so we send it here to avoid having to fetch it

### DIFF
--- a/test/unit/logic/community-voice.spec.ts
+++ b/test/unit/logic/community-voice.spec.ts
@@ -33,6 +33,8 @@ describe('Community Voice Logic', () => {
   let communityVoice: ICommunityVoiceComponent
   let logger: jest.Mocked<ReturnType<ILoggerComponent['getLogger']>>
   let mockCommunityVoiceChatCache: jest.Mocked<ICommunityVoiceChatCacheComponent>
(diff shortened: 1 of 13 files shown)


test/unit/logic/community-voice.spec.ts
```diff
@@ -249,14 +301,58 @@ describe('Community Voice Logic', () => {
...
+        })
+
+        describe('when placesApi fails', () => {
+          beforeEach(() => {
+            mockCatalystClient.getProfile.mockResolvedValue(createMockProfile(creatorAddress))
+            mockCommunitiesDb.getCommunityPlaces!.mockResolvedValue([{ id: 'place-1' }])
+            mockPlacesApi.getPlaces.mockRejectedValue(new Error('PlacesApi failed'))
+          })
+
+          it('should successfully start a community voice chat without positions when placesApi fails', async () => {
```
kevinszuchet, line 343: Would you mind removing the redundant `when placesApi fails` here?

```suggestion
          it('should successfully start a community voice chat without positions', async () => {
```

kevinszuchet (approved):

PR #250, opened 2025-08-15
feat: Allow global moderators to delete communities

kevinszuchet:


src/logic/community/communities.ts
```diff
@@ -297,13 +300,17 @@ export function createCommunityComponent(
     },
 
     deleteCommunity: async (id: string, userAddress: string): Promise<void> => {
+      // TODO: maybe we need to include a reason and store the reason in the database
```
aleortega, line 303: We talked about sending a new special notifications to owners of communities that have been deleted because of moderation (to be implemented in another PR), I think with this decision we won't need to store reason so we should be good.

aleortega (approved):
LGTM

PR #251, opened 2025-08-15
feat: Community compliance with ai

kevinszuchet:
### What
Implements automated content moderation for Decentraland communities using OpenAI's GPT-5 Nano to validate community names, descriptions, and thumbnails against our Code of Ethics.

### Why
- Ensures all community content aligns with Decentraland's values and policies
- Automatically flags violations of our Code of Ethics (violence, discrimination, illegal activities, etc.)
- Provides consistent, explainable compliance decisions with confidence scoring
- Reduces manual review workload for community submissions

### How
- **AI Compliance Component** (`ai-compliance.ts`): OpenAI integration with multimodal input support (text + images)
- **Compliance Validator** (`compliance-validator.ts`): Business logic layer that throws `CommunityComplianceError` for violations

### DIFF
--- a/src/controllers/middlewares/communities-errors.ts
+++ b/src/controllers/middlewares/communities-errors.ts
@@ -34,7 +34,7 @@ export async function communitiesErrorsHandler(
       }
     }
 
-    if (error instanceof InvalidCommunityRequestError || error instanceof AIComplianceError) {
+    if (error instanceof InvalidCommunityRequestError) {
       return {
         status: 400,
         body: {
@@ -48,11 +48,23 @@ export async function communitiesErrorsHandler(
       return {
         status: 400,
         body: {
+          error: 'Community not compliant',
           message: error.message,
           data: {
             issues: error.issues,
             warnings: error.warnings
           }
+        }
+      }
+    }
+
+    if (error instanceof AIComplianceError) {
+      return {
+        status: 400,
+        body: {
+          error: 'Community content validation unavailable',
+          message: error.message,
+          communityContentValidationUnavailable: true
         }
       }
     }
--- a/test/integration/create-community-controller.spec.ts
+++ b/test/integration/create-community-controller.spec.ts
@@ -362,6 +362,7 @@ test('Create Community Controller', async function ({ components, stubComponents
 
               expect(response.status).toBe(400)
               const body = await response.json()
+              expect(body.error).toBe('Community not compliant')
               expect(body.message).toContain("Community content violates Decentraland's Code of Ethics")
               expect(body.data.issues).toEqual(['Contains inappropriate language', 'Promotes violence'])
               expect(body.data.warnings).toEqual(['Content is borderline'])
@@ -370,14 +371,19 @@ test('Create Community Controller', async function ({ components, stubComponents
 
           describe('and AI compliance validation fails with AIComplianceError', () => {
             beforeEach(async () => {
-              stubComponents.communityComplianceValidator.validateCommunityContent.rejects(new AIComplianceError('AI compliance validation failed'))
+              stubComponents.communityComplianceValidator.validateCommunityContent.rejects(
+                new AIComplianceError('AI compliance validation failed')
+              )
             })
-            
+
             it('should respond with a 400 status code for AIComplianceError', async () => {
               const response = await makeMultipartRequest(identity, '/v1/communities', validBody)
 
               expect(response.status).toBe(400)
-              expect(await response.json()).toMatchObject({ message: 'AI compliance validation failed' })
+              const body = await response.json()
+              expect(body.error).toBe('Community content validation unavailable')
+              expect(body.message).toBe('AI compliance validation failed')
+              expect(body.communityContentValidationUnavailable).toBe(true)
             })
           })
 
--- a/test/integration/update-community-controller.spec.ts
+++ b/test/integration/update-community-controller.spec.ts
@@ -323,7 +323,7 @@ test('Update Community Controller', async function ({ components, stubComponents
                   },
                   'PUT'
                 )
-    
+
                 expect(response.status).toBe(200)
                 const body = await response.json()
                 expect(body.data.privacy).toBe(CommunityPrivacyEnum.Private)
@@ -333,11 +333,19 @@ test('Update Community Controller', async function ({ components, stubComponents
 
             describe('and the user is not the owner', () => {
               beforeEach(async () => {
-                await components.communitiesDb.updateMemberRole(communityId, identity.realAccount.address, CommunityRole.Moderator)
+                await components.communitiesDb.updateMemberRole(
+                  communityId,
+                  identity.realAccount.address,
+                  CommunityRole.Moderator
+                )
               })
 
               afterEach(async () => {
-                await components.communitiesDb.updateMemberRole(communityId, identity.realAccount.address, CommunityRole.Owner)
+                await components.communitiesDb.updateMemberRole(
+                  communityId,
+                  identity.realAccount.address,
+                  CommunityRole.Owner
+                )
               })
 
               it('should respond with a 401 status code', async () => {
@@ -382,6 +390,7 @@ test('Update Community Controller', async function ({ components, stubComponents
 
             expect(response.status).toBe(400)
             const body = await response.json()
+            expect(body.error).toBe('Community not compliant')
             expect(body.message).toContain("Community content violates Decentraland's Code of Ethics")
             expect(body.data.issues).toEqual(['Contains inappropriate language', 'Promotes violence'])
             expect(body.data.warnings).toEqual(['Content is borderline'])
@@ -400,13 +409,16 @@ test('Update Community Controller', async function ({ components, stubComponents
 
             expect(response.status).toBe(400)
             const body = await response.json()
+            expect(body.error).toBe('Community not compliant')
             expect(body.message).toContain("Community content violates Decentraland's Code of Ethics")
           })
         })
 
         describe('and AI compliance validation fails with AIComplianceError', () => {
           beforeEach(async () => {
-            stubComponents.communityComplianceValidator.validateCommunityContent.rejects(new AIComplianceError('AI compliance validation failed'))
+            stubComponents.communityComplianceValidator.validateCommunityContent.rejects(
+              new AIComplianceError('AI compliance validation failed')
+            )
           })
 
           it('should respond with a 400 status code for AIComplianceError', async () => {
@@ -420,7 +432,10 @@ test('Update Community Controller', async function ({ components, stubComponents
             )
 
             expect(response.status).toBe(400)
-            expect(await response.json()).toMatchObject({ message: 'AI compliance validation failed' })
+            const body = await response.json()
+            expect(body.error).toBe('Community content validation unavailable')
+            expect(body.message).toBe('AI compliance validation failed')
+            expect(body.communityContentValidationUnavailable).toBe(true)
           })
         })
       })


src/adapters/ai-compliance.ts
```diff
@@ -0,0 +1,191 @@
...
+        // Add image if provided
+        if (request.thumbnailBuffer) {
+          const dataUrl = `data:${request.thumbnailMime || 'image/png'};base64,${request.thumbnailBuffer.toString('base64')}`
+          userContent.push({ type: 'image_url', image_url: { url: dataUrl } })
+        }
+
+        // Production-optimized parameters
+        const completionParams = {
+          model,
+          // max_completion_tokens: 800, // Increased to allow reasoning + JSON output
```
aleortega, line 85: Is this comment expected?

src/adapters/ai-compliance.ts
```diff
@@ -0,0 +1,191 @@
...
+          hasImage: String(!!request.thumbnailBuffer),
+          model
+        })
+
+        const res = await openai.chat.completions.create({
+          ...completionParams,
+          messages
+        })
+
+        // Production logging - only what's useful
```
aleortega, line 112:
```suggestion
```

src/adapters/ai-compliance.ts
aleortega: This means we are not capping tokens used/money spent, right?

src/adapters/ai-compliance.ts
```diff
@@ -0,0 +1,191 @@
...
+
+        const content = res.choices[0]?.message?.content
+        if (!content) {
+          logger.error('No content in OpenAI response', {
+            requestId,
+            choices: JSON.stringify(res.choices),
+            finishReason: res.choices[0]?.finish_reason,
+            response: JSON.stringify(res, null, 2)
+          })
+          throw new Error('No content received from OpenAI API')
```
aleortega, line 131: I am not sure if we can trust OpenAPI to return correct answers all the time, at least until we test the approach and maybe refine the prompt.

I am thinking about the need to react to Open API failures, so in case it cannot return a proper answer for a certain community, we leave that community on-hold and ask for a manual action (slack-message, maybe?). Having said so, I'd try to throw a specific error instance from this module so then we can react over it in a higher-level component. Basically:

`if (error instaceof ComplianceError) setCommunityAsPendingAndSendSlackMessage()`

src/components.ts
```diff
@@ -203,6 +211,11 @@ export async function initComponents(): Promise<AppComponents> {
   })
   const communityOwners = createCommunityOwnersComponent({ catalystClient })
   const communityEvents = await createCommunityEventsComponent({ config, logs, fetcher, redis })
+
+  // AI Compliance components
+  const aiCompliance = await createAIComplianceComponent({ config, logs })
```
aleortega, line 216: Maybe we should only expose this component in higher environments (prevent execution of it in Zone, to prevent wasting tokens and instead of this component use a dummy one for dev), what do you think?

src/controllers/handlers/http/create-community-handler.ts
```diff
@@ -67,6 +67,19 @@ export async function createCommunityHandler(
       throw error
     }
 
+    if (error instanceof CommunityComplianceError) {
```
aleortega, line 70: Oh we already have a `CommunityComplianceError`, nice

src/controllers/handlers/http/update-community-handler.ts
```diff
@@ -66,6 +66,13 @@ export async function updateCommunityHandler(
       throw error
     }
 
+    if (error instanceof CommunityComplianceError) {
```
aleortega, line 69: What do you think about moving these checks to our error middleware?

src/logic/community/compliance-validator.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (!validationResult.isCompliant) {
+          logger.warn('Community content compliance validation failed', {
+            name,
+            issues: validationResult.issues.join(', '),
+            warnings: validationResult.warnings.join(', '),
+            confidence: validationResult.confidence,
+            duration
+          })
+
+          throw new CommunityComplianceError(
```
aleortega, line 54: Maybe we should distinguish between errors stating that the community did not pass the check and errors stating that the check could not be performed because of any reason

src/logic/community/compliance-validator.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        logger.error('Community content compliance validation error', {
+          name,
+          error: errorMessageOrDefault(error, 'Unknown error'),
+          duration
+        })
+
+        throw new CommunityComplianceError(
+          'Unable to validate community content compliance. Manual review required.',
```
aleortega, line 90: Nice, maybe we can use a boolean or a different error type to differentiate both cases (revision failed or revision could not be performed)

test/components.ts
```diff
@@ -186,6 +192,10 @@ async function initComponents(): Promise<TestComponents> {
   })
   const communityOwners = createCommunityOwnersComponent({ catalystClient })
   const communityEvents = await createCommunityEventsComponent({ config, logs, fetcher, redis })
+  const aiCompliance = await createAIComplianceComponent({ config, logs })
```
aleortega, line 195: We should use a mock here, right?

aleortega (commented):
Looking good, left some comments to start iterating over the solution but so far so goooood 🚀

src/adapters/ai-compliance.ts
```diff
@@ -0,0 +1,204 @@
...
+          typeof result.reasoning !== 'string'
+        ) {
+          logger.error('Invalid response structure', {
+            requestId,
+            result: JSON.stringify(result)
+          })
+          throw new AIComplianceError('Invalid response structure from OpenAI API')
+        }
+
+        const duration = Date.now() - startTime
```
aleortega, line 173: Would be great to report this duration as metric

src/adapters/ai-compliance.ts
```diff
@@ -0,0 +1,204 @@
...
+          requestId,
+          name: request.name,
+          isCompliant: String(result.isCompliant),
+          issuesCount: result.issues.length,
+          warningsCount: result.warnings.length,
+          confidence: result.confidence,
+          durationMs: duration
+        })
+
+        return result
```
aleortega, line 185: Also I'd report metrics to analyze successful requests and failing ones

src/adapters/communities-db.ts
```diff
@@ -468,9 +468,9 @@ export function createCommunitiesDBComponent(
     async createCommunity(community: CommunityDB): Promise<Community> {
       const id = randomUUID()
       const query = SQL`
-        INSERT INTO communities (id, name, description, owner_address, private, active)
-        VALUES (${id}, ${community.name}, ${community.description}, ${normalizeAddress(community.owner_address)}, ${community.private || false}, ${community.active || true})
-        RETURNING id, name, description, owner_address, private, active, created_at, updated_at
+        INSERT INTO communities (id, name, description, owner_address, private, active, needs_manual_review)
```
aleortega, line 471: This is for the future, right?

src/adapters/communities-db.ts
```diff
@@ -600,7 +601,7 @@ export function createCommunitiesDBComponent(
 
     async updateCommunity(
       communityId: string,
-      updates: Partial<Pick<CommunityDB, 'name' | 'description' | 'private'>>
+      updates: Partial<Pick<CommunityDB, 'name' | 'description' | 'private' | 'needs_manual_review' | 'active'>>
```
aleortega, line 604: We shouldn't allow updating `needs_manual_review` for now, right?
```suggestion
      updates: Partial<Pick<CommunityDB, 'name' | 'description' | 'private' | 'active'>>
```

src/adapters/feature-flags.ts
```diff
@@ -0,0 +1,81 @@
+import { IBaseComponent, START_COMPONENT, STOP_COMPONENT } from '@well-known-components/interfaces'
+import { AppComponents } from '../types'
+import { ApplicationName } from '@well-known-components/features-component'
+
+export enum FeatureFlag {
+  /**
+   * Feature flag to enable/disable AI compliance for communities.
+   *
+   * @example
+   * Variant value: "true" or "false"
```
aleortega, line 10: NIT: it won't have values on its variant, it will be just on or off

In FF terms, we are using `variant` concept to state that the FF besides being on or off it will also contain values

src/components.ts
```diff
@@ -152,6 +160,10 @@ export async function initComponents(): Promise<AppComponents> {
     pubsub,
     analytics
   })
+
+  const slackToken = await config.requireString('SLACK_BOT_TOKEN')
```
aleortega, line 164: Shall we create a fallback/mock component to use whenever `SLACK_BOT_TOKEN` isn't provided? (most likely to happen when running locally)

src/logic/community/communities.ts
```diff
@@ -248,11 +287,19 @@ export function createCommunityComponent(
         await communityPlaces.validateOwnership(placeIds, community.ownerAddress)
       }
 
+      const { needsManualReview, reason: manualReviewReason } =
```
aleortega, line 290: Sorry if we have talked about this, but aren't we delaying the process about approving communities manually? if that's the case also for creations, we should return a 400 status code instead of marking it as pending to be reviewed manually

src/logic/community/communities.ts
```diff
@@ -281,6 +328,21 @@ export function createCommunityComponent(
         }
       }
 
+      try {
```
aleortega, line 331: We are already catching inside of `sendManualReviewNotification` function we can remove this one

src/logic/community/compliance-validator.ts
```diff
@@ -0,0 +1,98 @@
...
+          hasThumbnail: String(!!thumbnailBuffer)
+        })
+
+        const validationResult = await aiCompliance.validateCommunityContent({
+          name,
+          description,
+          thumbnailBuffer
+        })
+
+        const duration = Date.now() - startTime
```
aleortega, line 47: Aren't we assessing duration in two different places?

src/utils/community-validation.ts
```diff
@@ -42,12 +42,16 @@ export async function validateCommunityFields(
   if (requireName || name !== undefined) {
     if (!name || typeof name !== 'string' || name.trim().length === 0) {
       throw new InvalidRequestError('Name must be a non-empty string')
+    } else if (name.length > 30) {
+      throw new InvalidRequestError('Name must be less or equal to 30 characters')
```
aleortega, line 46: Thanks a lot!

aleortega (commented):
Looking good to me, only thing pending to solve is to decide how should we treat this feature in lower environments such as DEV, we can discuss it tomorrow during the sync!

src/adapters/communities-db.ts
kevinszuchet: I will revert all the manual review related changes. It's confusing. In the future we can see these commits if we need them

src/adapters/feature-flags.ts
kevinszuchet: I left the old comment and only changed the values, nice catch, thanks!

src/components.ts
kevinszuchet: I only moved this component higher up in the code. The token is included in the .env.default file, so the require statement won't fail. However, you're correct that if the flow you're testing locally needs to send messages to Slack, it will eventually encounter an error. 

Perhaps we can address this in a separate pull request.

PR #257, opened 2025-08-18
feat: adjust permissions on community call actions

juanmahidalgo:
This PR changes the logic to:
- for kicking, kick the user no matter what, if the community is private/public it doesn't matter.
- for promoting/demoting: always if public. If private, it needs to be part of the community and not be banned.

### DIFF
--- a/src/logic/community-voice/validation.ts
+++ b/src/logic/community-voice/validation.ts
@@ -1,0 +1,39 @@
+import { CommunityRole } from '../../types/entities'
+import { UserNotCommunityMemberError } from './errors'
+import { ICommunitiesDatabaseComponent } from '../../types/components'
+import { Community } from '../community/types'
+
+/**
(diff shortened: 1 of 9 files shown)


src/logic/community-voice/validation.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param targetUserAddress - Target user address
+ * @returns Promise<void> - Throws error if validation fails
+ */
+export async function validateCommunityVoiceChatTargetUser(
+  communitiesDb: ICommunitiesDatabaseComponent,
+  community: Community,
+  communityId: string,
+  targetUserAddress: string
+): Promise<void> {
+  // Validation logic based on community type
```
kevinszuchet, line 20: Can we do an early return if the community is not private to avoid the if/else? I think it could simplify the function

PR #259, opened 2025-08-19
feat: endpoint to retrieve community invites

aleortega:
# Add Community Invites Endpoint
Implements a new endpoint to retrieve communities where an inviter is a member but an invitee is not, enabling users to see which communities they can invite others to.

### New Endpoint
- **GET** `/v1/members/{address}/invites` - Returns communities available for invitation

#### Response structure

```typescript
[
 {
   id: string
   name: string 
 }
]
```


src/controllers/handlers/http/get-community-invites-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+  const logger = logs.getLogger('get-community-invites-handler')
+
+  try {
+    const inviterAddress = verification!.auth.toLowerCase()
+
+    if (inviterAddress === inviteeAddress) {
+      throw new InvalidRequestError('Users cannot invite themselves')
+    }
+
+    if (!EthAddress.validate(inviterAddress) || !EthAddress.validate(inviteeAddress)) {
```
braianj, line 33: Just a question: isn't `inviterAddress` validated with the signedFetchMiddleware?

braianj (approved):
🎉

src/controllers/handlers/http/get-community-invites-handler.ts
aleortega: I checked and it seems it isn't validating addresses explicitly with this function!

PR #263, opened 2025-08-26
fix: only admins can change community name and places can be removed only if you own it

aleortega:
This PR adds restrictions on Community updates so:
- Name can only be changed by Owner
- Places can only be removed by Owners or by Moderators who own them

### DIFF
--- a/src/logic/community/communities.ts
+++ b/src/logic/community/communities.ts
@@ -342,6 +342,9 @@ export function createCommunityComponent(
       }
 
       await communityRoles.validatePermissionToEditCommunity(communityId, userAddress)
+      if (updates.name && updates.name.trim() !== existingCommunity.name.trim()) {
+        await communityRoles.validatePermissionToEditCommunityName(communityId, userAddress)
+      }
 
       if (Object.keys(updates).length === 0) {
         return {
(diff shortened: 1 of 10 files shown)


src/logic/community/communities.ts
```diff
@@ -342,6 +342,9 @@ export function createCommunityComponent(
       }
 
       await communityRoles.validatePermissionToEditCommunity(communityId, userAddress)
+      if (updates.name && updates.name.trim() !== existingCommunity.name.trim()) {
```
kevinszuchet, line 345: Does it matter if both names are equal if we lowercase them?

test/unit/logic/community-places.spec.ts
```diff
@@ -38,6 +38,16 @@ describe('Community Places Component', () => {
...
     mockPlacesApi = createPlacesApiAdapterMockComponent({}) as jest.Mocked<IPlacesApiComponent>
+    
+    // Re-setup mockLogs after creating mocks
+    mockLogs.getLogger.mockReturnValueOnce({
+      log: jest.fn(),
+      debug: jest.fn(),
+      error: jest.fn(),
+      info: jest.fn(),
+      warn: jest.fn()
+    })
```
kevinszuchet, line 49: Would you mind exposing a `createMockLogs` function in the `mocks/component/logs.ts` file if we need to redefine the mock functions?

kevinszuchet (approved):
LGTM

test/unit/logic/community-places.spec.ts
aleortega: Sure thing! I'll refactor 👼

PR #269, opened 2025-08-29
feat: More optimizations

kevinszuchet:


src/adapters/ai-compliance.ts
```diff
@@ -120,6 +106,26 @@ export async function createAIComplianceComponent(
       const startTime = Date.now()
       const requestId = `${Date.now()}-${Math.random().toString(36).substr(2, 9)}`
 
+      // Build a content hash to memoize identical inputs
+      const imageHash = request.thumbnailBuffer
+        ? crypto.createHash('sha256').update(request.thumbnailBuffer).digest('hex')
+        : 'no-img'
+      const cacheKey = `ai-compliance:${model}:${crypto
```
aleortega, line 113: NIT:
Move cache key generation to a new function

src/components.ts
```diff
@@ -213,7 +215,7 @@ export async function initComponents(): Promise<AppComponents> {
   const communityEvents = await createCommunityEventsComponent({ config, logs, fetcher, redis })
 
   // AI Compliance components
-  const aiCompliance = await createAIComplianceComponent({ config, logs, featureFlags, metrics })
+  const aiCompliance = await createAIComplianceComponent({ config, logs, featureFlags, metrics, memoryCache })
```
aleortega, line 218: Shouldn't we use Redis instead, so the cache doesn't get flushed when the service restarts? Also, a good refactor would be:
```typescript
const memoryStorage = redisHost ? createRedis(...) : createInMemoryCache(...)
```

aleortega (approved):
LGTM, left a comment about using Redis instead of in-memory cache

PR #286, opened 2025-09-02
feat: notify about users join whenever a request is accepted

aleortega:
Whenever a Community Request is accepted (and a member joins a community), the client should be notified through WS so it can update its chat state.


kevinszuchet (approved):
LGTM!

PR #299, opened 2025-10-01
fix: do not show private communities voice chats statuses to all

juanmahidalgo:
Do not show the Community Voice chat status of private communities if the user is not prart of when asking for communities

### DIFF
--- a/src/logic/community/utils.ts
+++ b/src/logic/community/utils.ts
@@ -73,11 +73,8 @@ export const toCommunityWithUserInformationAndVoiceChat = (
 ): CommunityWithUserInformationAndVoiceChat => {
   const baseResult = toCommunityWithUserInformation(community, profilesMap)
 
-  // If the community is private and the user is not a member, don't include voice chat status
-  // This prevents non-members from knowing if a private community has an active voice chat
-  const shouldIncludeVoiceChat = !(
-    community.privacy === CommunityPrivacyEnum.Private && community.role === CommunityRole.None
-  )
+  const shouldIncludeVoiceChat =
+    community.privacy !== CommunityPrivacyEnum.Private || community.role !== CommunityRole.None
 
   return {
     ...baseResult,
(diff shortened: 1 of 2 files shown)


test/unit/logic/community/utils.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+        expect(result.voiceChatStatus).toEqual(mockVoiceChatStatus)
+      })
+
+      it('should include voiceChatStatus for owners', () => {
+        const community = { ...mockCommunity, privacy: CommunityPrivacyEnum.Public, role: CommunityRole.Owner }
+        const result = toCommunityWithUserInformationAndVoiceChat(community, profilesMap, mockVoiceChatStatus)
+
+        expect(result.voiceChatStatus).toEqual(mockVoiceChatStatus)
+      })
+    })
```
kevinszuchet, line 66: Could you move the descriptions of the different test contexts to the statements in the `describe` blocks?

For example:

```js
  describe('and the user is a member', () => {
      it('should include voiceChatStatus', () => {
        const community = { ...mockCommunity, privacy: CommunityPrivacyEnum.Public, role: CommunityRole.Member }
        const result = toCommunityWithUserInformationAndVoiceChat(community, profilesMap, mockVoiceChatStatus)

        expect(result.voiceChatStatus).toEqual(mockVoiceChatStatus)
      })
  })
```

src/logic/community/utils.ts
```diff
@@ -71,9 +72,16 @@ export const toCommunityWithUserInformationAndVoiceChat = (
...
+  // If the community is private and the user is not a member, don't include voice chat status
+  // This prevents non-members from knowing if a private community has an active voice chat
+  const shouldIncludeVoiceChat = !(
+    community.privacy === CommunityPrivacyEnum.Private && community.role === CommunityRole.None
+  )
+
   return {
     ...baseResult,
-    voiceChatStatus
+    voiceChatStatus: shouldIncludeVoiceChat ? voiceChatStatus : null
```
kevinszuchet, line 84: NIT: wdyt if we change it to `shouldHideVoiceChatStatus` to remove the `!` at the beginning. Then we can do `shouldHideVoiceChatStatus ? null : voiceChatStatus`.

It sounds clearer to me.

kevinszuchet (approved):
LGTM

PR #324, opened 2025-10-29
feat: Unlist Communities

kevinszuchet:
When creating or editing a community, we will have an additional toggle `Visibility: All | Unlisted`. When a community is `Unlisted`, it will not appear in the community browser nor the search results, but it will be visible in the Requests and Invitations area.


test/integration/update-community-controller.spec.ts
```diff
@@ -234,6 +242,36 @@ test('Update Community Controller', async function ({ components, stubComponents
...
+                })
+              })
+
+              afterEach(async () => {
+                await components.communitiesDb.updateCommunity(communityId, {
+                  private: false
+                })
+              })
+
+              it('should not switch the community to public', async () => {
```
LautaroPetaccio: (I'm taking into consideration that this test is executed under the context that a description exists)
As there are multiple variations of the properties that can be updated, it might be better in terms of maintenance to check these variations with unit tests. We could focus on cases that test that each property was properly updated.

PR #329, opened 2025-10-31
feat: Add editors choice flagging

kevinszuchet:


src/controllers/handlers/http/update-community-partially-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+    params: { id: communityId },
+    verification,
+    request
+  } = context
+
+  const logger = logs.getLogger('update-community-partially-handler')
+  const userAddress = verification!.auth.toLowerCase()
+
+  try {
+    const body: UpdateCommunityPartiallyBody = await request.json()
```
aleortega: May we can use this shared component to handle these validations:
https://github.com/decentraland/core-components/blob/main/components/schema-validator/src/component.ts

src/controllers/handlers/http/update-community-partially-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const hasAnyUpdate = Object.keys(body).length > 0
```
aleortega: Currently only `editorsChoice` is supported, right? I'd refactor this validation to check just that since all additional properties would be ignored

src/logic/community/communities.ts
```diff
@@ -519,6 +519,28 @@ export function createCommunityComponent(
       return updatedCommunity
     },
 
+    updateEditorChoice: async (communityId: string, userAddress: EthAddress, editorsChoice: boolean): Promise<void> => {
+      const community = await communitiesDb.getCommunity(communityId, userAddress)
+      if (!community) {
+        throw new CommunityNotFoundError(communityId)
+      }
+
+      const globalModerators =
```
aleortega: Shall we normalize these addresses just in case?

PR #333, opened 2025-10-31
feat: Track community metrics and calculate scoring

kevinszuchet:


src/logic/community/ranking.ts
```diff
@@ -0,0 +1,88 @@
...
+      const communitiesMetrics = await communitiesDb.getAllCommunitiesWithRankingMetrics()
+
+      if (communitiesMetrics.length === 0) {
+        logger.info('No active communities found for ranking calculation')
+        return
+      }
+
+      // Check thumbnails for all communities in a single batch
+      const communityIds = communitiesMetrics.map(({ communityId }) => communityId)
+      const thumbnails = await communityThumbnail.getThumbnails(communityIds)
```
LautaroPetaccio: Would checking the thumbnail for each community slow down the query's response?

src/logic/community/ranking.ts
```diff
@@ -0,0 +1,88 @@
...
+          }
+          const score = await calculateRankingScore(metrics)
+
+          logger.info('Calculated ranking score for community', {
+            communityId: communityMetrics.communityId,
+            score
+          })
+
+          // Update ranking score
+          await communitiesDb.updateCommunity(communityMetrics.communityId, { ranking_score: score })
```
LautaroPetaccio: What if we either:
- Store the `ranking_score` in the `community_ranking_metrics` when inserting or updating the table, instead of inserting it in the community's table. This would require a LEFT JOIN with the metrics.
- Compute the `ranking_score` when getting the communities by doing a LEFT JOIN with the `community_ranking_metrics` table.

Having the metrics data on a single table would simplify all the process of updating and computing them, specially the part of being consistent with the data in both tables. I would compute them separately as it's done here only if there's a performance issue.
What do you think about this approach?

PR #336, opened 2025-11-05
feat: New notifications, broadcaster refactor, author profile when responding create post

kevinszuchet:


src/logic/community/broadcaster.ts
```diff
@@ -154,51 +187,107 @@ export function createCommunityBroadcasterComponent(
...
-      await publishToOwnersAndModerators(event as CommunityRequestToJoinReceivedEventReducedMetadata)
-    } else if (event.subType === Events.SubType.Community.DELETED_CONTENT_VIOLATION) {
-      // Notify the owner
-      await sns.publishMessage(event)
-
-      // Send a different notification to the rest of the community members
-      const communityDeletedEvent = {
+  /**
+   * Broadcasts to all members except a specific excluded address
+   * Looks for excludedAddress in metadata, or falls back to common field names like authorAddress, senderAddress, etc.
```
aleortega:
```suggestion
```

src/logic/community/broadcaster.ts
```diff
@@ -154,51 +187,107 @@ export function createCommunityBroadcasterComponent(
...
+
+  /**
+   * Creates the broadcasting registry with all strategies
+   * @returns {BroadcastingRegistry} Map of event subTypes to their broadcasting handlers
+   */
+  function createBroadcastingRegistry(): BroadcastingRegistry {
+    const registry = new Map<Events.SubType.Community, BroadcastingEventHandler>()
+
+    // Events that should be broadcasted to all members
+    registry.set(Events.SubType.Community.DELETED, broadcastToAllMembers)
```
aleortega: Since the Owner is the only role that could delete the Community, could we double-check with product if they should receive this `DELETED` event?

Mostly because if we can avoid delivering it to them, we could handle the `DELETED_CONTENT_VIOLATION` differently and maybe more meaningful, here is my idea:
- Whenever `DELETED_CONTENT_VIOLATION` happens, we could trigger two events:
  - `COMMUNITY_DELETED`: to be notified to all members but owners (if Product agree with us)
  -  `DELETED_CONTENT_VIOLATION`: to be notified to only owners

This way we could set:
```
registry.set(Events.SubType.Community.DELETED, broadcastToAllMembersButOwners)
registry.set(Events.SubType.Community.DELETED_CONTENT_VIOLANTION, broadcastToOwnersAndModerators) // or broadcastToOwner
```

src/logic/community/members.ts
```diff
@@ -119,6 +121,27 @@ export async function createCommunityMembersComponent(
     }
 
     await communitiesDb.transferCommunityOwnership(communityId, targetAddress)
+    setImmediate(async () => {
+      const community = await communitiesDb.getCommunity(communityId)
```
aleortega: Two doubts here:
* Shouldn't we check if the community exists even before executing `await communityRoles.validatePermissionToTransferOwnership(communityId, updaterAddress, targetAddress)`? Another solution would to make that function to return community_id value in case it exists so we prevent going once again to the database
* Shouldn't we update the broadcaster to also notify this event `CommunityOwnershipTransferredEvent`?

kevinszuchet: The entry point is the `updateMemberRole` function, which checks whether the community exists before calling the `transferOwnership` function, which is not publicly exposed.

I thought the same, but it's weird to call a broadcaster to send a notification to only one user. WDYT?

src/logic/community/broadcaster.ts
```diff
@@ -154,51 +187,107 @@ export function createCommunityBroadcasterComponent(
...
-
-      // Send a different notification to the rest of the community members
-      const communityDeletedEvent = {
+  /**
+   * Broadcasts to all members except a specific excluded address
+   * Looks for excludedAddress in metadata, or falls back to common field names like authorAddress, senderAddress, etc.
+   * @param {BroadcastableEvent} event - The event to broadcast
+   * @throws {Error} If excluded address or community ID is not found in metadata
+   */
+  async function broadcastToAllMembersButExcluded(event: BroadcastableEvent): Promise<void> {
```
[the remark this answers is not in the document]
kevinszuchet: I'm not sure what you mean here. The goal is to broadcast the notification while avoiding notifications to the user who initiated the action (only the post author ATM)

src/logic/community/broadcaster.ts
kevinszuchet: I really like this idea

aleortega (approved):
LGTM

PR #339, opened 2025-11-12
feat: fetch profiles from content-server instead of lambdas

aleortega:
This PR updates how the service fetches profiles from Catalyst. Instead of fetching the profiles from Lambda's client, which performs ownership checks behind the scenes, it will use the Content client from now on.

The rationale behind this change is that the service only needs to validate that the profile exists and grabs its set name; it doesn't use the items from the profile, so there is no need to perform ownership checks.

### DIFF
--- a/src/adapters/catalyst-client.ts
+++ b/src/adapters/catalyst-client.ts
@@ -115,6 +115,21 @@ export async function createCatalystClient({
 
       const response = await retry(executeClientRequest, retries, waitTime)
       validProfiles = response.map((entity) => extractMinimalProfile(entity?.metadata)).filter(Boolean) as Profile[]
+      const minimalProfiles = validProfiles.map(extractMinimalProfile).filter(Boolean) as Profile[]
+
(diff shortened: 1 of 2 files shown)


src/adapters/catalyst-client.ts
```diff
@@ -27,6 +27,26 @@ export async function createCatalystClient({
...
 
+  function getContentClientOrDefault(contentServerUrl?: string): ContentClient {
+    return createContentClient({ fetcher, url: contentServerUrl ?? loadBalancer })
+  }
+
+  function rotateContentServerClient<T>(
+    executeClientRequest: (client: ContentClient) => Promise<T>,
+    contentServerUrl?: string
+  ) {
+    const contentServers = shuffleArray(getCatalystServersFromCache(contractNetwork)).map((server) => server.address)
```
LautaroPetaccio: Should we remove the `contentServerUrl` passed as parameter from the `contentServers`?

LautaroPetaccio (approved):

PR #350, opened 2025-12-19
feat: Community voice chat started notification through RPC and Notifications Service

kevinszuchet:


src/logic/community/broadcaster.ts
```diff
@@ -222,6 +242,85 @@ export function createCommunityBroadcasterComponent(
...
+
+    const onlineSubscribers = subscribersContext.getSubscribersAddresses()
+
+    if (onlineSubscribers.length === 0) {
+      return
+    }
+
+    // Get online members from the community, excluding specified addresses at database level
+    const excludedAddresses = options?.excludeAddresses?.map((addr) => addr.toLowerCase())
+    const onlineMemberAddresses = await getOnlineCommunityMembersAddresses(
```
aleortega: NIT:
This function, in reality, only grabs members based on address, right? Could we reuse the get members function we already have in the database and remove this one? maybe we can enhance the original one to expect options to exclude addresses, or we could exclude the address before calling the function

src/logic/community/broadcaster.ts
```diff
@@ -222,6 +242,85 @@ export function createCommunityBroadcasterComponent(
...
+    if (onlineMemberAddresses.length === 0) {
+      return
+    }
+
+    const memberBatches = createMemberBatches(onlineMemberAddresses)
+
+    await sns.publishMessages(
+      memberBatches.map((batch, i) => ({
+        ...event,
+        key: `${event.key}-batch-${i + 1}`,
```
aleortega: NIT:
The key can be a generic one, since the event will arrive once per user address.

DOUBT:
What's `event.key` here? asking because if it is something like `<communityId>`, all voice chats starting will have same same key, meaning that only the first voice chat will be notified, afterwards notification would detect that there is already a notification for that key and that address and exclude them on retrieval.

Even if we add the `batch-<number>`, that index will be the same.

I think we already have a unique event.keys at this point but since I am not sure I leave this comment 👼

aleortega (approved):
LGTM!

PR #374, opened 2026-02-02
feat: support search communities returning minimal data

kevinszuchet:
## Summary
Adds a lightweight community search mode to the existing GET /v1/communities endpoint via a new minimal=true query parameter. This enables fast community lookups (e.g., for autocomplete) without fetching the full community data.

### Key changes:
- Extended GET /v1/communities to support minimal=true mode that returns only id and name
- Added optimized prefix-matching search query for fast community name lookups
- Requires authentication and a minimum 3-character search query when using minimal mode
- Unlisted communities are only searchable by their members

### Key Decisions
- Uses ILIKE for prefix and word-boundary matching rather than full-text search for simplicity
- Mandatory Signed Fetch auth allowing only authenticated searches

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -114,9 +114,10 @@ paths:
         
         **Minimal Mode (minimal=true):**
         When `minimal=true` is set, authentication is required and only the `search` parameter is used:
-        - Returns only community `id` and `name` for fast lookups
+        - Returns community `id`, `name`, `membersCount`, and `privacy` for fast lookups
         - Requires `search` parameter with minimum 3 characters
         - Maximum 50 results per request
+        - Supports pagination with `limit` and `offset` parameters
         - Other filter parameters are ignored
         
         **Visibility Rules (Minimal Mode):**
@@ -154,8 +155,9 @@ paths:
             type: boolean
             default: false
           description: |
-            When `true`, returns only community `id` and `name` for fast lookups.
+            When `true`, returns community `id`, `name`, `membersCount`, and `privacy` for fast lookups.
             Requires authentication and `search` parameter with minimum 3 characters.
+            Supports pagination with `limit` (max 50) and `offset` parameters.
             Other filter parameters (`onlyMemberOf`, `onlyWithActiveVoiceChat`, `roles`) are ignored.
         - name: onlyMemberOf
           in: query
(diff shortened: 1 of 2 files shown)


src/adapters/communities-db.ts
```diff
@@ -1282,6 +1282,61 @@ export function createCommunitiesDBComponent(
...
+            c.unlisted = false
+            OR EXISTS (
+              SELECT 1 FROM community_members cm
+              WHERE cm.community_id = c.id AND cm.member_address = ${normalizedUserAddress}
+            )
+          )
+      `
+
+      const result = await pg.query<{ count: string }>(query)
+      return parseInt(result.rows[0]?.count ?? '0', 10)
```
LautaroPetaccio: WDYT about something like this to avoid duplicating the query?
```suggestion
    async searchCommunities(
      search: string,
      options: { userAddress: EthAddress; limit: number; offset: number }
    ): Promise<{ results: Array<{ id: string; name: string }, total: number}>> {
      const { userAddress, limit, offset } = options
      const normalizedUserAddress = normalizeAddress(userAddress)

      // Optimized prefix matching query:
      // - Matches names starting with the search term
      // - Also matches words in the middle of the name (after a space)
      // - Public and Private communities are always searchable
      // - Unlisted communities are only searchable by their members
      const mainQuery = SQL`SELECT c.id, c.name`
      const countQuery = SQL`SELECT COUNT(*) as count`
      const body = SQL`
        FROM communities c
        WHERE c.active = true
          AND (c.name ILIKE ${search + '%'} OR c.name ILIKE ${'% ' + search + '%'})
          AND (
            c.unlisted = false
            OR EXISTS (
              SELECT 1 FROM community_members cm
              WHERE cm.community_id = c.id AND cm.member_address = ${normalizedUserAddress}
            )
          )
      `
      mainQuery.append(body)
      countQuery.append(body)
      
      const mainQuery = countQuery.append(`ORDER BY c.name ASC
        LIMIT ${limit}
        OFFSET ${offset}`)

      const [results, count] = await Promise.all(pg.query<{ id: string; name: string }>(mainQuery), pg.query<{ count: string }>(countQuery))
      return {
        results: results.rows,
        count: parseInt(count.rows[0]?.count ?? '0', 10)
      }
    }
```

PR #379, opened 2026-02-09
feat: Fire even when someone leaves a community

### LINKED ISSUE #378: Publish `MEMBER_LEFT` event on SNS when a user leaves a community

kevinszuchet:
Closes #378

### DIFF
--- a/src/logic/community/broadcaster.ts
+++ b/src/logic/community/broadcaster.ts
@@ -10,7 +10,8 @@ import {
   Events,
   CommunityPostAddedEvent,
   CommunityOwnershipTransferredEvent,
-  CommunityVoiceChatStartedEvent
+  CommunityVoiceChatStartedEvent,
+  CommunityMemberLeftEvent
 } from '@dcl/schemas'
 import { AppComponents, CommunityRole } from '../../types'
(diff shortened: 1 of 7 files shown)


src/logic/community/broadcaster.ts
```diff
@@ -10,7 +10,8 @@ import {
   Events,
   CommunityPostAddedEvent,
   CommunityOwnershipTransferredEvent,
-  CommunityVoiceChatStartedEvent
+  CommunityVoiceChatStartedEvent,
+  CommunityMemberLeftEvent
```
aleortega: I think we are not using CommunityMemberRemovedEvent at all at the moment, right? I thought that event was being reported for this case as well. Maybe we could use that same event for both purposes: trigger when the member is removed by an admin/mod and trigger when the member leaves (and differentiate both cases based on a intention flag [self, removed])

This is not a strong-opinion, but I am thinking about a potential service which need to react to members removal/leaves from community, with this current approach we should hook the new service to both kind of events instead to a single one. What do you think?

Again, this is not a strong-opinion but I like to thinking about events about "global" things that have happened in a specific domain, in this case a better name could be CommunityMemberLeave (prop flag|reason: self||removal)

This is not a blocker, I am leaving the approve anyway!

PR #406, opened 2026-03-24
fix: resolve OOM crashes with heap limit, TTL-based subscribers, and SCAN

kevinszuchet:
## Summary
- **community-voice-cache**: Replace `KEYS` with `SCAN` to avoid blocking Redis, and use `mGet` instead of N individual `get` calls (N round-trips → 1)
- **localstack**: Pin to `4.12.0` — the `latest` tag broke CI with exit code 55 after the `2026.03.0` release on March 23
- **instrument.ts**: Restore `--inspect` flag (remove hardcoded `--max-old-space-size`; V8 auto-sizes the heap from the container's cgroup memory limit)

## Context
The OOM crashes (`Ineffective mark-compacts near heap limit`) are primarily caused by insufficient container memory — the service had **no memory/CPU reservation configured**, so V8 auto-sized to a tiny heap. The infrastructure fix is in [definitions#1431](https://github.com/decentraland/definitions/pull/1431) (adds 1024 MB prod / 512 MB dev).

This PR addresses the **application-side** improvements: more efficient Redis usage and removing the hardcoded heap override so V8 can auto-size correctly from the new container limits.


src/logic/community-voice/community-voice-cache.ts
```diff
@@ -108,7 +108,12 @@ export function createCommunityVoiceChatCacheComponent({
 
   async function getActiveCommunityVoiceChats(): Promise<CachedCommunityVoiceChat[]> {
     try {
-      const keys = await redis.client.keys(`${CACHE_PREFIX}*`)
+      // Use SCAN instead of KEYS to avoid blocking Redis and loading all keys at once
+      const keys: string[] = []
+      for await (const key of redis.client.scanIterator({ MATCH: `${CACHE_PREFIX}*`, COUNT: 100 })) {
```
LautaroPetaccio, line 113: Maybe we can extend this iteration loop to retrieve the `CachedCommunityVoiceChat` from redis using `mget` to do less round-trips to redis. WDYT?

LautaroPetaccio (approved):

src/logic/community-voice/community-voice-cache.ts
kevinszuchet: The rationale was to avoid doing big blocker calls to Redis, that's why the iterator is in batches. Maybe the mGet can work below when actually getting the values related to the keys. WDYT? Maybe in a different PR

src/logic/community-voice/community-voice-cache.ts
```diff
@@ -108,26 +108,18 @@ export function createCommunityVoiceChatCacheComponent({
...
-              error: isErrorWithMessage(error) ? error.message : 'Unknown error'
-            })
-            return null
-          }
-        })
-      )
-
-      return chats.filter((chat): chat is CachedCommunityVoiceChat => chat !== null)
+      const chats = await redis.mGet<CachedCommunityVoiceChat>(keys)
+      return chats.filter((chat) => chat.isActive)
```
kevinszuchet, line 122: Changed this many gets for a multi get. FYI @LautaroPetaccio

PR #407, opened 2026-04-15
fix: prevent OOM crashes from subscriber leak and duplicate subscriptions

kevinszuchet:
## Summary

Fixes the recurring OOM crash (exit code 134, ~510 MB heap after ~62 hours) by addressing all root causes identified from crash logs and code analysis:

### Critical fixes (commit 1)
- **Cleanup bug** (`ws-handler.ts`): Three independent cleanup steps were in a single `try/catch`. If `transport.close()` threw (uWS throws on closed sockets), `detachUser()` and `emitter.clear()` were skipped — permanently leaking the subscriber's mitt emitter and all registered async generators. Extracted each step into its own safe function.
- **Duplicate subscriptions** (`updates.ts` + `subscribers-context.ts`): Clients were calling the same stream RPC twice per connection (confirmed by double "Cleaning up" logs). Each duplicate created another async generator with a 1000-item value queue, doubling memory per connection. Added per-(address, event) deduplication guard.
- **No heap limit** (`Dockerfile`): V8 auto-sized to the full 1024 MB container with no headroom for native allocations. Added `NODE_OPTIONS=--max-old-space-size=768` (75% of prod container memory), overridable at deploy time.

### Hardening (commit 2)
- **Cork wrapping** (`uws.routes.ts`): All HTTP response writes now happen inside `res.cork()`, eliminating the uWS cork warnings that appeared under GC pressure.
- **Subscriber metrics**: New gauges for `subscribers_local_count`, `subscribers_generators_count`, `subscribers_redis_set_size`, and `subscribers_stale_cleaned` counter — reported every 30s.
- **Stale subscriber reconciliation**: Every 5 minutes, cross-references local subscribers against active WS connections and removes orphans. Configurable via `SUBSCRIBER_RECONCILIATION_INTERVAL_MS`.
- **sScan instead of sMembers**: `getSubscribersAddresses()` now uses `sScanIterator` to avoid loading the entire Redis SET at once. Warns when the set exceeds 10,000 entries.

## Test plan

- [x] TypeScript compiles clean (`npx tsc --noEmit`)
- [x] All 1533 unit tests pass (2 pre-existing failures from uWS native module incompatibility on macOS, unrelated)
- [x] subscribers-context + updates tests pass (66/66 + 17/17)
- [ ] Deploy to staging and monitor new `subscribers_*` metrics
- [ ] Verify no more "Cleaning up ... subscription (x2)" duplicate logs
- [ ] Verify subscriber count stays bounded over 24+ hours
- [ ] Verify no more uWS cork warnings
- [ ] Verify reconciliation logs appear every 5 min with `removed=0` under normal conditions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,10 @@ ENV COMMIT_HASH=${COMMIT_HASH:-local}
 ARG CURRENT_VERSION=Unknown
 ENV CURRENT_VERSION=${CURRENT_VERSION:-Unknown}
 
+# Set V8 heap limit to 75% of container memory (prod=1024MB, so 768MB).
+# Override at deploy time via environment variable for non-prod (e.g. 384 for 512MB containers).
+ENV NODE_OPTIONS="--max-old-space-size=768"
+
 WORKDIR /app
 COPY --from=builderenv /app /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,10 +31,6 @@ ENV COMMIT_HASH=${COMMIT_HASH:-local}
 ARG CURRENT_VERSION=Unknown
 ENV CURRENT_VERSION=${CURRENT_VERSION:-Unknown}
 
-# Set V8 heap limit to 75% of container memory (prod=1024MB, so 768MB).
-# Override at deploy time via environment variable for non-prod (e.g. 384 for 512MB containers).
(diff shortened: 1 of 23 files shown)


LautaroPetaccio (approved):
Looks great! 🚀

LautaroPetaccio (approved):

PR #429, opened 2026-07-02
fix: harden WS transport, subscriptions, RPC startup and backpressure

LautaroPetaccio:
## Summary

A review of the WebSocket / RPC / subscription / backpressure paths surfaced a cluster of related issues. Under sustained backpressure the transport would *cancel its own backoff, burn its retries, silently drop a message (stalling the RPC stream), spam unhandled rejections, and leave the socket open*. This PR fixes that cluster plus several independent issues found alongside it.

All changes are covered by unit tests — **107 suites / 1645 tests green**, typecheck and lint clean.

## Bugs fixed

- **Unhandled promise rejections** (`UWebSocketTransport`): `@dcl/rpc` calls `sendMessage` fire-and-forget, but the returned `fp-future` had no catch handler, so every reject (drop, send error, connection-closed) surfaced as an `unhandledRejection`. Production runs with `--unhandled-rejections=strict`, so this was a real process-crash vector, not just noise. Now a no-op catch is attached at enqueue time; awaiting callers still observe the rejection.
- **Silent drop → protocol corruption**: after max retries the message was dropped and skipped. The `@dcl/rpc` stream protocol is strictly ack-based — a dropped stream frame leaves the pump suspended forever, and a dropped unary response hangs the client call. Now the transport emits `error`, the RPC server tears the transport down (rejecting everything queued), and the client reconnects cleanly.
- **Backoff defeated by concurrent sends**: a new `send()` during a pending backoff cancelled the timer and immediately re-sent the backpressured head, burning all retries within a burst. Now the queue isn't re-kicked while a backoff is pending, and the uWS `drain` event is wired through so retries happen the moment the socket actually drains.
- **Zombie sockets**: when the RPC layer closed the transport (e.g. queue overflow) the uWS socket was never ended, leaving the client with a live, ping-alive connection whose messages went nowhere. The transport `close` listener now ends the socket (`1011`).
- **`start()` didn't await pub/sub subscriptions** (`rpc-server`): an inner `forEach` returned `undefined`, so `Promise.all` resolved immediately and the subscribe promises floated. Combined with `subscribeToChannel` swallowing errors, an instance could boot with no subscriptions. Now `flatMap`/`map` awaits the real promises and `subscribeToChannel` rethrows so boot fails loud.
- **Fabricated ONLINE presence** (`updates`): friendship-accept announced both parties as ONLINE unconditionally. Now it announces a friend as ONLINE only if they're actually connected (checked against `peersStats`).
- **`ws.end()` on a closed socket** in the auth-failure path could throw and escape the async message handler as an unhandled rejection — now guarded, matching the existing auth-timeout guard.

## Improvements

- **Snapshot/live gap** (friend-connectivity subscription): the initial snapshot is now delivered *after* the live listener registers (via a new optional hook on `handleSubscriptionUpdates`), so updates emitted during the snapshot queries are queued rather than lost.
- **Attach race**: `authenticateUser` re-checks `isConnected` after transport creation before attaching the user, and cleanup clears `isConnected` before closing the transport so the zombie-socket guard is precise on the normal close path instead of relying on uWS throwing.
- **Per-connection reconciliation** (`subscribers-context`): the sweep now also removes leaked `wsConnectionId`s whose socket is gone while the address still has a live connection — previously such a leak kept a user "online" forever. Adds `wsPool.getConnectionIds()` and a `subscribers_stale_connections_cleaned` metric.
- **Connection cap**: optional `WS_MAX_CONCURRENT_CONNECTIONS` (unset = unlimited, so no behavior change until configured), surfaced as a typed `WsPoolFullError` mapped to the existing `1013 Try again later` close.
- **Idempotent `transport.close()`**, dead `connect` emit removed, `ws_messages_sent` counted on actual outbound sends, and stream metrics recorded via try/finally so client-closed streams are no longer uncounted.

## Notes for review / deploy

- **`ws_messages_sent` changed meaning**: it now counts outbound transport frames (including RPC control frames) rather than inbound forwarded messages — adjust any dashboard/alert on it.
- **Boot now fails loud on a transient Redis blip during `start()`** (intended: silent no-subscriptions was worse; the orchestrator restarts).
- Watch **`ws_backpressure_events{result="max_retries"}`** after deploy — each hit is now a deliberate disconnect where before it was a silent message loss.

## Testing

- `yarn test:unit` — 107 suites / 1645 tests pass.
- Typecheck (`tsc --noEmit`) and `yarn lint:check` clean.
- Integration suite (`yarn test:integration`, needs docker) not run in this environment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LautaroPetaccio:
Two subtle behavioral notes from a final review pass, for reviewers:

1. **Duplicate subscriptions no longer receive the initial snapshot.** Previously, a connection re-subscribing to `subscribeToFriendConnectivityUpdates` while its first stream was still active got the online-friends snapshot again followed by a dead (never-yielding) stream, since the duplicate guard (#407) blocks the second live stream. Now the snapshot runs inside `handleSubscriptionUpdates` *after* the duplicate guard, so a duplicate gets a cleanly-completed empty stream with no snapshot — and no redundant DB/registry queries for clients stuck in a re-subscribe loop. Any client relying on re-subscribe-for-fresh-snapshot was already broken by the guard; this just makes the rejection cheaper.

2. **Accept-time ONLINE notifications degrade to "skip" during a stats outage.** `peersStats.getConnectedPeers()` swallows source failures into an empty list, so if both archipelago and worlds stats are unreachable at the moment a friendship is accepted, the ONLINE notifications are skipped silently rather than fabricated (the old behavior announced ONLINE unconditionally). Since peer-tracking only publishes on status *transitions*, an already-online friend won't be re-announced until their next transition — acceptable and preferable to false presence, but worth knowing. Documented in the code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

LautaroPetaccio:
Addressed the review comments in 73de67f:

1. **`ws.end()` guard in the open handler** — Added, matching the other two sites. Worth noting it's not reachable today: no `await` precedes it in the callback, so the socket can't have closed mid-handler — but the guard future-proofs against an `await` landing in that try block and keeps the pattern uniform.

2. **Max-retries branch relying on the external error listener** — Took the stronger option: kept the `emit('error')` teardown path (and documented the contract in a comment), *and* added a defensive fallback — if no listener closed the transport during the emit, the poisoned head message is rejected and dropped so the queue provably can't spin on it. Added a regression test asserting a subsequent message still sends after max retries.

3. **`ws_messages_sent` help text** — Updated the help string to "Number of outbound WebSocket messages accepted by the socket (sent or buffered)" plus a declaration comment noting the counting-point change (includes RPC control frames; previously counted inbound forwarded messages). Since the help string changes the exposed Prometheus HELP line, dashboards keyed on the metric *name* are unaffected.

4. **Double `detachUser` on close paths** — Agreed, no change: it's intentional defense-in-depth. The transport close listener needs it for the RPC-initiated path, and `cleanupConnection` needs its own independent call so a throw in `transport.close()` can't skip user detachment (each cleanup step is deliberately isolated). `removeConnection` treats untracked connections as a no-op, so the second call is harmless by design.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

